PR #16, opened 2017-07-28
Allow daemonizing the process

dasch:
Allow daemonizing the Racecar process. When executing `racecar` with the `--daemonize` flag, the process will fork off and do the daemonic dance in order to liberate itself from its terminal master. The process id (PID) of the new process will be written to a (configurable) file, which will allow tools to identify the process and send it signals. The `racecarctl` tool gets two new commands, `status` and `stop`.

Based on http://codeincomplete.com/posts/ruby-daemons/

This is done in order to support users who do not want to use process managers such as Runit or God. A Capistrano recipe could be added to automate the process management.

#### Tasks

- [x] Should we keep a PID file per consumer class and have `racecarctl` take a consumer class argument? For example, `racecarctl stop EchoConsumer`?

### DIFF
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -1,6 +1,8 @@
 require "optparse"
 require "logger"
+require "fileutils"
 require "racecar/rails_config_file_loader"
+require "racecar/daemon"
 
 module Racecar
   module Cli
@@ -17,7 +19,9 @@ module Racecar
         end
 
         Racecar::Config.variables.each do |variable|
-          opt_name = "--" << variable.name.to_s.gsub("_", "-") << " VALUE"
+          opt_name = "--" << variable.name.to_s.gsub("_", "-")
+          opt_name << " VALUE" unless variable.boolean?
+
           desc = variable.description || "N/A"
 
           if variable.default
@@ -25,7 +29,13 @@ module Racecar
           end
 
           opts.on(opt_name, desc) do |value|
-            Racecar.config.decode(variable.name, value)
+            if variable.boolean?
+              # Boolean switches are automatically mapped to true/false.
+              Racecar.config.set(variable.name, value)
+            else
+              # Other CLI params need to be decoded into values of the correct type.
+              Racecar.config.decode(variable.name, value)
+            end
           end
         end
 
@@ -63,7 +73,28 @@ module Racecar
       end
 
       $stderr.puts "=> Wrooooom!"
-      $stderr.puts "=> Ctrl-C to shutdown consumer"
+
+      if Racecar.config.daemonize
+        daemon = Daemon.new(File.expand_path(Racecar.config.pidfile))
+
+        daemon.check_pid
+
+        $stderr.puts "=> Starting background process"
+        $stderr.puts "=> Writing PID to #{daemon.pidfile}"
+
+        daemon.suppress_input
+
+        if Racecar.config.logfile.nil?
+          daemon.suppress_output
+        else
+          daemon.redirect_output(Racecar.config.logfile)
+        end
+
+        daemon.daemonize!
+        daemon.write_pid
+      else
+        $stderr.puts "=> Ctrl-C to shutdown consumer"
+      end
 
       processor = consumer_class.new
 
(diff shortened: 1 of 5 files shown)


lib/racecar/cli.rb
```diff
@@ -1,13 +1,21 @@
...
 
 module Racecar
   module Cli
     def self.main(args)
+      daemonize = false
+
       parser = OptionParser.new do |opts|
         opts.banner = "Usage: racecar MyConsumer [options]"
 
+        opts.on("-d", "--daemonize", "Run the consumer daemonized in the background") do
```
bquorning: No `--pid` option?

bquorning: Ah no, the pidfile can be set in the configuration file.

lib/racecar/daemon.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    def pidfile?
+      !pidfile.nil?
+    end
+
+    def daemonize!
+      exit if fork
+      Process.setsid
+      exit if fork
+      Dir.chdir "/"
```
bquorning: perhaps also `STDIN.reopen "/dev/null"`?

lib/racecar/daemon.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def redirect_output(logfile)
+      FileUtils.mkdir_p(File.dirname(logfile), :mode => 0755)
```
bquorning: A hashrocket, such nostalgia. Rubocop would have told you to use `mode: 0755` here. I won’t…

lib/racecar/daemon.rb
```diff
@@ -0,0 +1,91 @@
...
+    rescue Errno::ESRCH
+      :dead
+    rescue Errno::EPERM
+      :not_owned
+    end
+
+    def write_pid
+      begin
+        File.open(pidfile, ::File::CREAT | ::File::EXCL | ::File::WRONLY) do |f|
+          f.write("#{Process.pid}")
```
bquorning: I prefer `Process.pid.to_s` over `"#{Process.pid}"`, but whatever.

lib/racecar/daemon.rb
```diff
@@ -0,0 +1,91 @@
...
+    def stop!
+      Process.kill("TERM", pid)
+    end
+
+    def pid_status
+      return :exited if pid.nil?
+      return :dead if pid == 0
+
+      # check process status
+      Process.kill(0, pid)
```
bquorning: What does this do? Which `trap` is invoked?

lib/racecar/cli.rb
dasch: Fixed.

lib/racecar/cli.rb
```diff
@@ -45,7 +53,26 @@ def self.main(args)
...
+        if Racecar.config.logfile.nil?
+          daemon.suppress_output
+        else
+          daemon.redirect_output(Racecar.config.logfile)
+        end
+
+        daemon.daemonize!
+        daemon.write_pid
+      else
+        $stderr.puts "=> Ctrl-C to shutdown consumer"
```
dasch: Operator information is printed to stderr while runtime logging happens on stdout. That way, doing `racecar MyConsumer > racecar.log` would still show important info in the terminal.

lib/racecar/ctl.rb
```diff
@@ -17,6 +18,33 @@ def self.main(args)
...
+      daemon = Daemon.new(pidfile)
+
+      if daemon.running?
+        puts "running (PID = #{daemon.pid})"
+      else
+        puts daemon.pid_status
+      end
+    end
+
+    def stop(args)
```
dasch: This is invoked by the CLI.

lib/racecar/daemon.rb
dasch: :+1:

lib/racecar/daemon.rb
dasch: 🤗

lib/racecar/daemon.rb
dasch: Will update the comment – it just raises an exception if the process doesn't exist.

lib/racecar/daemon.rb
dasch: :+1:

Issue #12, opened 2017-07-24
Batch processing API


filiptepper:
We're considering using `racecar` for some of our Kafka consumers, but our use case requires processing whole batches of messages.

Would you be interested in a contribution that would allow that kind of behavior, built on top of `each_batch` method?

dasch:
I've considered how best to add batch processing APIs. Since I'd like to keep things easy for newcomers, I think simply implementing a `process_batch` method rather than `process` method would probably work best – Racecar can detect which method the consumer class implements and act accordingly.

Can you give me more detail on your use case?

filiptepper:
We use Kafka consumers to write data to different data stores, including Amazon's DynamoDB and Elasticsearch. To improve performance, resource usage and costs we batch multiple Kafka messages to one (or more, depending on how many messages consumer receiver) query.

`process_batch` would be my preferred way of processing as well.

Do you think it would possible to easily return a success / failure from `process_batch` to indicate whether consumer offset position should be updated?

dasch:
@filiptepper there's already an implicit indication: if your method raises an exception, the position won't be advanced; if it doesn't, the position _will_ be advanced. This has worked well until now.

filiptepper:
@dasch Perfect!

I've forked to repository to add some basic `process_batch` and test it with our use case. Will get back to you.

dasch:
:+1:

dasch:
Fixed by #17.

PR #17, opened 2017-08-01
Batch processing

filiptepper:
Per discussion in issue #12.

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -53,7 +53,7 @@ module Racecar
             processor.process_batch(batch)
           end
         else
-          raise NotImplementedError, "Consumer class must implement process or process_match method", caller
+          raise NotImplementedError, "Consumer class must implement process or process_batch method", caller
         end
       rescue Kafka::ProcessingError => e
         @logger.error "Error processing partition #{e.topic}/#{e.partition} at offset #{e.offset}"
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -44,8 +44,16 @@ module Racecar
       end
 
       begin
-        consumer.each_message(max_wait_time: config.max_wait_time) do |message|
-          processor.process(message)
+        if processor.respond_to?(:process)
+          consumer.each_message(max_wait_time: config.max_wait_time) do |message|
+            processor.process(message)
+          end
+        elsif processor.respond_to?(:process_batch)
+          consumer.each_batch(max_wait_time: config.max_wait_time) do |batch|
+            processor.process_batch(batch)
+          end
+        else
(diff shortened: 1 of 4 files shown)


spec/runner_spec.rb
```diff
@@ -16,11 +16,37 @@ def on_message(&block)
...
+    @processor_queue = []
+  end
+
+  def on_message(&block)
+    @processor_queue << block
+    self
+  end
+
+  def process_batch(messages)
+    @messages += messages
```
dasch: Actually, ruby-kafka doesn't pass an array of messages to the block in `#each_batch` – it passes a [FetchedBatch](http://www.rubydoc.info/gems/ruby-kafka/Kafka/FetchedBatch) object that contains a `#messages` method as well as other metadata that's actually useful. I think `#process_batch` should be called with that batch object.

dasch: It would be a good idea to add an example batch consumer to `examples/` – you can run it with

    ruby -Ilib exe/racecar --require examples/batch_consumer BatchConsumer

filiptepper: You're absolutely right about it - I didn't notice that until a bit later, after I created this PR. Will address this and come back with some updates.

Thanks!

filiptepper:
Just pushed changes you requested, let me know if there's anything else that needs fixing!

lib/racecar/runner.rb
```diff
@@ -44,8 +44,16 @@ def run
...
+        if processor.respond_to?(:process)
+          consumer.each_message(max_wait_time: config.max_wait_time) do |message|
+            processor.process(message)
+          end
+        elsif processor.respond_to?(:process_batch)
+          consumer.each_batch(max_wait_time: config.max_wait_time) do |batch|
+            processor.process_batch(batch)
+          end
+        else
+          raise NotImplementedError, "Consumer class must implement process or process_batch method", caller
```
dasch: Why pass `caller` here?

filiptepper:
Just dropped it. It doesn't really belong here.

dasch:
@filiptepper I've released this as part of `v0.3.0.beta1` – can you try it out and see if it works for you?

filiptepper:
@dasch Actually, we're already running it in production since Tuesday (based off my fork with latest updates from `zendesk/master`).

Later today / maybe tomorrow I'll try switching to `v0.3.0.beta1` release and get back to you.

filiptepper:
Bumped our production instance to `v0.3.0.beta1`, so far so good!

dasch:
Awesome! Unless you report a problem here within the next few days I'll release v0.3.0 next week.

filiptepper:
Heads up - still running on beta, still no issues.

dasch:
@filiptepper awesome! I've just released v0.3.0 final.

PR #20, opened 2017-08-17
call `stop` on consumer class when trapping signals

filiptepper:
As we're rewriting our consumers we've noticed that there's one case we can't handle with racecar right now. Some of our consumer do some cleanup when a consumer is being shut down. This PR would allow us to have the same behaviour in racecar.

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -2,10 +2,15 @@ require "kafka"
 
 module Racecar
   class Runner
-    attr_reader :processor, :config, :logger
+    attr_reader :processor, :config, :logger, :consumer
 
     def initialize(processor, config:, logger:)
       @processor, @config, @logger = processor, config, logger
+    end
+
+    def stop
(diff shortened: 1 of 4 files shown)


lib/racecar/runner.rb
```diff
@@ -2,12 +2,17 @@
...
   class Runner
-    attr_reader :processor, :config, :logger
+    attr_reader :processor, :config, :logger, :consumer
 
     def initialize(processor, config:, logger:)
       @processor, @config, @logger = processor, config, logger
     end
 
+    def stop
+      processor.stop if processor.respond_to?(:stop)
```
dasch: Could we call this callback e.g. `teardown`? I think that better communicates that it's a callback that allows tearing down resources, not a method that actually stops anything.

filiptepper: I needed this so much, couldn't fine a better name than `stop`, `teardown` is so much more obvious. :)

filiptepper:
Made some adjustments & added notes to `README`.

PR #23, opened 2017-08-21
Fix typo in README

jbdietrich:
Sorry, my inner pedant got the better of me.


bquorning:
Yeah, that seams right. :+1:

PR #27, opened 2017-09-14
Use KingKonf for configuration

dasch:


mensfeld:
@dasch  wouldn't it be better to use dry ecosystem libraries for that? Seems redundant.

dasch:
@mensfeld I couldn't find any library that worked the way I wanted it to: simple, flat key/value structure, with support for ENV vars and nice usability.

PR #33, opened 2017-10-05
Better rails loading

staugaard:
I improved the rails app and config loading by doing this:

1. Don't swallow too many `LoadError`
2. Only load the configuration file if it is present.


bquorning:
@dasch Could you release a new version, too?

PR #34, opened 2017-10-06
Defer loading of config until after rails had a chance to set ENV vars

staugaard:
Before we would initialize the config very early in the CLI, so rails apps would not get a chance to set ENV vars in their initializers.


dasch:
I’m a bit skeptical of an app mutating its own env, but I’m fine with this change 👍

PR #39, opened 2017-10-20
Support older ActiveSupport versions

FlorianDoublet:
For some old versions of ActiveSupport, ActiveSupport::Logger does not exist.


dasch:
👍

PR #40, opened 2017-10-23
Allow configuring Racecar with Ruby

dasch:
* Automatically loads `config/racecar.rb` if it exists.
* Adds a nicer syntax for configuring in Ruby.

Now you can just do this:

```ruby
# config/racecar.rb
Racecar.configure do |config|
  config.session_timeout = 10

  # Provide a default value if e.g. no ENV var has been provided
  # and there's no Racecar default:
  config.group_id ||= "some-group"
end
```

### DIFF
--- a/lib/racecar.rb
+++ b/lib/racecar.rb
@@ -15,6 +15,10 @@ module Racecar
     @config ||= Config.new
   end
 
+  def self.configure
+    yield config
+  end
+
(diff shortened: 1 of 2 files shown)


dasch:
cc @staugaard

staugaard:
Nice!

PR #42, opened 2017-11-28
Fix log_level

sparrovv:
`logger.level=` expects INT, not a String


dasch:
Hmm, the string support must be in a more recent version of ruby then.

dasch:
Thanks!

PR #43, opened 2017-11-30
Instrument using ActiveSupport::Notifications

dasch:


lib/racecar.rb
```diff
@@ -27,7 +34,15 @@ def self.logger=(logger)
     @logger = logger
   end
 
+  def self.instrumenter
+    ActiveSupport::Notifications
+  rescue NameError
```
bquorning: Did you consider `require 'active_support/notifications' ; rescue LoadError`? That’s the pattern I usually see people use.

dasch: Good point.

PR #44, opened 2017-12-05
Pass in the SASL options

sparrovv:
Looks like the SASL options weren't passed in.


mensfeld:
Would be good to coordinate this with: https://github.com/zendesk/delivery_boy/issues/13 as some people use DB with Rc together.

PR #46, opened 2017-12-11
Fix `consumer_class` instrumentation field

p-lambert:
This PR fixes the `consumer_class` instrumentation param provided to the `instrumenter`. 

All event subscribers should now receive a `consumer_class` field that looks like `FooConsumer` instead of `#<FooConsumer:0x007ff7bf7cc3b0>`.


dasch:
Thanks!

PR #49, opened 2017-12-31
Pass ssl_ca_cert_file_path config to Kafka.new in runner.rb

webandtech:
It looks like the `ssl_ca_cert_file_path` option was missed in the runner - this fixes the issue.


dasch:
Thanks! ❤️

Issue #29, opened 2017-09-18
Support for producing Kafka messages


dasch:
A lot of our use cases are centered around _transforming_ streams of Kafka messages and writing them back to another topic, typically enriched with data from our Rails app. Currently, Racecar provides no direct support for producing messages. Adding direct support would allow the delivery of messages to be optimized.

aviflax:
@dasch [DeliveryBoy](https://github.com/zendesk/delivery_boy) is the library that Zendesk has released for producing to Kafka. You can combine them together to do stream processing, or you can use [ruby-kafka](https://github.com/zendesk/ruby-kafka) which is the more flexible, more powerful, lower-level library that’s the foundation for them both. HTH!

dasch:
Thank you! I'm the author of both libraries, so I'm aware of that 😁 

I was thinking about an integrated solution that coupled the delivery step to the batch processing of Racecar.

aviflax:
Doh! Sorry about that! I guess I just wasn’t paying attention 🤷‍♂️

In that case, BTW, +1! Just today I put together an example of doing stream processing with ruby-kafka — I couldn’t figure out how to use Racecar in a script, outside of Rails — and some kind of built-in simple stream processing feature would be excellent for beginners!

Cheers!

webandtech:
@dasch We also do this. I created a PR. Hope it's useful!

PR #50, opened 2018-01-02
Support for producing messages from #process / #process_batch return value

webandtech:
[#29] This PR adds support for a common use case where a consumer needs to process messages from one topic and produce messages on another. Examples include data transformation or "responding" to another service that a message has been processed.

#### Producing responses

Use the optional `responds_with:` option to specify a topic to produce messages to. The return value from the `#process` or `#process_batch` method will be the message produced. If the return value is `nil`, no message will be produced.

`response_partition_key:` may be used to specify a partition key or a method pointer (symbol) to determine the partition key used when producing the response.

```ruby
class AddUserIdConsumer < Racecar::Consumer
  subscribes_to "visitors", responds_with: "visitors_user_id", response_partition_key: :transformed_partition_key
  
  def process(message)
    data = JSON.parse(message.value)
    user = User.find_by(email: data['email'])]
    return unless user
    
    data['user_id'] = user.id
    @partition_key = user.account_id
    data.to_json
  end
  
  def transformed_partition_key
    @partition_key
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,38 @@ end
 
 An important detail is that, if an exception is raised while processing a batch, the _whole batch_ is re-processed.
 
+#### Producing responses
+
+A common use case is to process messages from one topic, and then "respond" to another topic, such as when transforming
+data or logging the result of the processed message.
+
+Use the optional `responds_with:` option to specify a topic to produce messages to. The return value from the 
+`#process` or `#process_batch` method will be the message produced. If the return value is `nil`, no message
+will be produced.
+
+The `response_partition_key:` may be used to specify a partition key or a method pointer (symbol). When using a method,
+the message or batch will be passed to the method and it's return used as the partition key for the message produced.
+
+```ruby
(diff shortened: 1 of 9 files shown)


dasch:
This isn't quite the API I had in mind:

1) I'd rather avoid adding a dependency on DeliveryBoy, as it's not really necessary – ruby-kafka has all the required bits.
2) Racecar should be less "magical" than this, and more explicit. It's not meant to be a stream processing framework with fancy functional APIs.
3) Producing to several different topics is a very common use case.

Let me see if I can write a PR that implements my ideas this week.

dasch:
#51 is my initial stab at this.

webandtech:
No worries. I will close this PR for now. Perhaps I’ll refactor it to be a small self-contained extension in its own gem. What do you think about Racecar extensions?

I went both ways on the DeliveryBoy dependency but chose to include it because I felt like there would be duplication across the 2 gems increasing maintenance overhead over time to keep Racecar’s producer api up to date.

PR #51, opened 2018-01-02
Allow producing messages

### LINKED ISSUE #29: Support for producing Kafka messages

dasch:
This change allows consumers to produce messages. The messages will be delivered automatically after `#process` returns, in the case of single-message processing, or after `#process_batch` returns, in the case of batch processing.

The API works like this:

```ruby
class DoublingConsumer < Racecar::Consumer
  subscribes_to "numbers"

  def process(message)
    value = Integer(message.value)

    # The #produce method buffers up a message to be delivered to Kafka.
    # Racecar handles the actual delivery after #process returns.
    produce value * 2, topic: "doubled"
  end
end
```

Closes #29.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,13 @@ The framework is based on [ruby-kafka](https://github.com/zendesk/ruby-kafka), w
 2. [Usage](#usage)
     1. [Creating consumers](#creating-consumers)
     2. [Running consumers](#running-consumers)
-    3. [Configuration](#configuration)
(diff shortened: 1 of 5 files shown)


webandtech:
:thumbsup

PR #54, opened 2018-01-26
Change the default max_wait_time to 1 second

tjwp:
In https://github.com/zendesk/ruby-kafka/pull/434, the default `max_wait_time` in ruby-kafka was changed to 1 second.

This make the corresponding change for racecar, since the default here overrides the ruby-kafka default.


dasch:
Thanks!

PR #58, opened 2018-02-01
[Readme update] Use Ruby class instead of Rails class

berkos:
Use Ruby class instead of Rails class as this is a plain Ruby class.


dasch:
Thanks!

PR #69, opened 2018-03-19
Document how to use Kubernetes to deploy Racecar

dasch:


inprogs:
I am not sure if this work was forgotten or just waiting for new update, anyway it is valid piece of information. I want to try this with google cloud. :+1:  for this

PR #75, opened 2018-05-09
Use duration config vars

dasch:


mensfeld:
So I guess we won't support neither 10.5s or 100ms, right?

dasch:
The latest version of KingKonf supports e.g. `10s 500ms` and `100ms`.

Issue #92, opened 2018-09-14
Different clusters for consumers


Jakanapes:
Is there support for having consumers read from different clusters?

Can the brokers be overridden in the consumer class?

dasch:
Since each consumer runs in its own process, you can start each with a different set of brokers:

```sh
$ RACECAR_BROKERS=kafka1.cluster1,kafka2.cluster1 racecar --daemonize Cluster1Consumer
$ RACECAR_BROKERS=kafka1.cluster2,kafka2.cluster2 racecar --daemonize Cluster1Consumer
```

PR #93, opened 2018-09-14
allow consumers to set brokers

Jakanapes:
Let consumers set their own brokers with `self.brokers = ['broker1', 'broker2']`

### DIFF
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -162,6 +162,8 @@ module Racecar
     end
 
     def load_consumer_class(consumer_class)
+      self.brokers = consumer_class.brokers || self.brokers
+
       self.group_id = consumer_class.group_id || self.group_id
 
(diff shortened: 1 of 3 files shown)


dasch:
I'm not sure I like this API – the brokers are not an aspect of the consumer, they're information related to the environment the consumer is running in. What is the use case for this?

Jakanapes:
https://github.com/zendesk/racecar/issues/92

I see your response there, so it's good to know there is an existing workaround.

We have a local cluster we read from and a 3rd party cluster, totally different consumers.

The way we were thinking of it was since the client creation is abstracted away, that the consumer can control which environment it chooses to read from.   This would also make it easier on future devs, with the connection information in the racecar domain, instead of somewhat hidden in a Procfile or systemd script

dasch:
I would want to separate the environment from the domain code.

PR #96, opened 2018-10-14
Fix graceful shutdown

### LINKED ISSUE #86: log writing failed. can't be called from trap context
When a racecar consumer is running and I send a SIGTERM to it. It prints the following:

Jun 11 18:55:49 ip-10-212-33-14 systemd[1]: Stopping App client consumer. Instance=1....
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 bash[9370]: log writing failed. can't be called from trap context
Jun 11 18:55:49 ip-10-212-33-14 systemd[1]: Stopped App client consumer. Instance=1..
Not always but sometimes.

I think this maybe because we can not use logger in signal handler https://bugs.ruby-lang.org/issues/7917, but ruby-kafka is doing this. If you follow trap signal handling by racecar you will finally land in a code in ruby-kafka that uses [logger.info
](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/broker_pool.rb#L36)

Solution may be to create a new thread like karafka. But I don't like this method. 

Ruby-kafka ticket ref: https://github.com/zendesk/ruby-kafka/issues/598

Version of Ruby: 2.3.1
Version of Kafka: 0.11.0.1
Version of ruby-kafka: 0.5.2
Version of racecar: 0.3.7

JHK:
This fixes #86 with the hack to create a thread and join it again that
was mentioned in this issue.

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -10,8 +10,10 @@ module Racecar
     end
 
     def stop
-      processor.teardown
-      consumer.stop unless consumer.nil?
+      Thread.new do
+        processor.teardown
+        consumer.stop unless consumer.nil?
+      end.join
     end
 
     def run


dasch:
Thanks!

Issue #57, opened 2018-02-01
Over Reads on a Topic


GeminPatel:
Hi,

I started a consumer subscribing to an empty topic. Then with extreme speed dumped 37869 messages in a that clean topic. My consumer side by side started fetching. Looking at the log shows that it over processed. It processed till 37904 and then reset back to the correct one. Why did this happen? And this time my count of processes also reduced to 37756 instead of 37869.

Log:
```

D, [2018-02-01T19:20:55.952360 #13490] DEBUG -- : Marking users/0:37990 as processed
D, [2018-02-01T19:20:55.953482 #13490] DEBUG -- : Marking users/0:37868 as processed
D, [2018-02-01T19:20:55.956459 #13490] DEBUG -- : Marking users/0:37869 as processed
D, [2018-02-01T19:20:55.957973 #13490] DEBUG -- : Marking users/0:37870 as processed
D, [2018-02-01T19:20:55.959402 #13490] DEBUG -- : Marking users/0:37871 as processed
D, [2018-02-01T19:20:55.960880 #13490] DEBUG -- : Marking users/0:37872 as processed
D, [2018-02-01T19:20:55.962321 #13490] DEBUG -- : Marking users/0:37873 as processed
D, [2018-02-01T19:20:55.963735 #13490] DEBUG -- : Marking users/0:37874 as processed
D, [2018-02-01T19:20:55.965176 #13490] DEBUG -- : Marking users/0:37875 as processed
D, [2018-02-01T19:20:55.966594 #13490] DEBUG -- : Marking users/0:37876 as processed
D, [2018-02-01T19:20:55.968016 #13490] DEBUG -- : Marking users/0:37877 as processed
D, [2018-02-01T19:20:55.969427 #13490] DEBUG -- : Marking users/0:37878 as processed
D, [2018-02-01T19:20:55.970775 #13490] DEBUG -- : Marking users/0:37879 as processed
D, [2018-02-01T19:20:55.972231 #13490] DEBUG -- : Marking users/0:37880 as processed
D, [2018-02-01T19:20:55.975846 #13490] DEBUG -- : Marking users/0:37881 as processed
D, [2018-02-01T19:20:55.978254 #13490] DEBUG -- : Marking users/0:37882 as processed
D, [2018-02-01T19:20:55.980070 #13490] DEBUG -- : Marking users/0:37883 as processed
D, [2018-02-01T19:20:55.981492 #13490] DEBUG -- : Marking users/0:37884 as processed
[...]

PR #55, opened 2018-01-26
Expose offset_retention_time configuration for consumers

tjwp:
This change exposes the `offset_retention_time` from the ruby-kafka Consumer API (https://github.com/zendesk/ruby-kafka/pull/316) for Racecar consumers.

The `offset_retention_time` can be set globally or for an individual consumer class.

PR #97, opened 2018-10-16
Change ruby-kafka to rdkafka-ruby 

JHK:
This PR will change the underlying library to the C-based [rdkafka-ruby](https://github.com/appsignal/rdkafka-ruby/). It is still work in progress as some things are not updated yet:

- [x] Tests
- [x] Documentation on configuration
- [x] Documentation on breaking changes
- [x] Instrumentation of rdkafka-ruby (https://github.com/appsignal/rdkafka-ruby/issues/54)


README.md
```diff
@@ -114,7 +114,7 @@ This is useful to do any one-off work that you wouldn't want to do for each and
 When a consumer is started for the first time, it needs to decide where in each partition to start. By default, it will start at the _beginning_, meaning that all past messages will be processed. If you want to instead start at the _end_ of each partition, change your `subscribes_to` like this:
 
 ```ruby
-subscribes_to "some-topic", start_from_beginning: false
+subscribes_to "some-topic", config: { "auto.offset.reset" => "largest" }
```
dasch: I think this is one of those cases where it makes sense to have a wrapper around the low-level config while still providing the low level stuff as a fallback. So I'd keep `start_from_beginning` but add `config`.

lib/racecar/config.rb
```diff
@@ -10,20 +10,29 @@ class Config < KingKonf::Config
     desc "A string used to identify the client in logs and metrics"
     string :client_id, default: "racecar"
 
-    desc "How frequently to commit offset positions"
-    float :offset_commit_interval, default: 10
```
dasch: Is there a similar mechanism in rdkafka for setting the offset commit frequency?

dasch: Yeah, `auto.commit.interval.ms`. I would keep this config.

JHK: Search for `auto.commit.interval.ms` in https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md (sadly cannot deep link). Defaults to 5 seconds.

lib/racecar/config.rb
```diff
@@ -34,81 +43,18 @@ class Config < KingKonf::Config
...
-    float :connect_timeout, default: 10
-
-    desc "How long to wait when trying to communicate with a Kafka broker"
-    float :socket_timeout, default: 30
-
-    desc "How long to allow the Kafka brokers to wait before returning messages"
-    float :max_wait_time, default: 1
-
-    desc "The maximum size of message sets returned from a single fetch"
-    integer :max_bytes, default: 10485760
```
dasch: So of these are available in rdkafka – I'd like to keep these high-level config keys and just copy to the matching ones in rdkafka.

README.md
```diff
@@ -237,8 +237,6 @@ end
 The consumers will checkpoint their positions from time to time in order to be able to recover from failures. This is called _committing offsets_, since it's done by tracking the offset reached in each partition being processed, and committing those offset numbers to the Kafka offset storage API. If you can tolerate more double-processing after a failure, you can increase the interval between commits in order to better performance. You can also do the opposite if you prefer less chance of double-processing.
 
 * `offset_commit_interval` – How often to save the consumer's position in Kafka. Default is every 10 seconds.
-* `offset_commit_threshold` – How many messages to process before forcing a checkpoint. Default is 0, which means there's no limit. Setting this to e.g. 100 makes the consumer stop every 100 messages to checkpoint its position.
-* `offset_retention_time` - How long committed offsets will be retained. Defaults to the broker setting.
```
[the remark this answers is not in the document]
dasch: `offset_commit_threshold` is a ruby-kafka thing, but `offset_retention_time` should be supported – it's part of the protocol.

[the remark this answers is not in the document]
JHK: Okay I tried to dig a bit deeper there. It seems like the underlying C library [librdkafka](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md) does not expose offset retention as configuration option. Looking at the current [Apache Kafka Documentation](https://kafka.apache.org/documentation/) I can only find it as broker configuration value.

Personally I don't think we would need that option to be set from the consumer side. Therefore I would vote for ditching it to be able to make progress.

FlorianDoublet:
Good initiative 👍 !

lib/racecar/config.rb
```diff
@@ -66,14 +66,53 @@ class Config < KingKonf::Config
     desc "The log level for the Racecar logs"
     string :log_level, default: "info"
 
+    desc "Protocol used to communicate with brokers"
+    symbol :security_protocol, allowed_values: %i{plaintext ssl sasl_plaintext sasl_ssl} + %w{plaintext ssl sasl_plaintext sasl_ssl}
```
dasch: Why allow strings here?

lib/racecar/config.rb
```diff
@@ -144,15 +183,39 @@ def on_error(&handler)
...
+      security_config["ssl.keystore.password"] = ssl_keystore_password unless ssl_keystore_password.nil?
+      security_config["sasl.mechanism"] = sasl_mechanism unless sasl_mechanism.nil?
+      security_config["sasl.kerberos.service.name"] = sasl_kerberos_service_name unless sasl_kerberos_service_name.nil?
+      security_config["sasl.kerberos.principal"] = sasl_kerberos_principal unless sasl_kerberos_principal.nil?
+      security_config["sasl.kerberos.kinit.cmd"] = sasl_kerberos_kinit_cmd unless sasl_kerberos_kinit_cmd.nil?
+      security_config["sasl.kerberos.keytab"] = sasl_kerberos_keytab unless sasl_kerberos_keytab.nil?
+      security_config["sasl.kerberos.min.time.before.relogin"] = sasl_kerberos_min_time_before_relogin unless sasl_kerberos_min_time_before_relogin.to_i.zero?
+      security_config["sasl.username"] = sasl_username unless sasl_username.nil?
+      security_config["sasl.password"] = sasl_password unless sasl_password.nil?
+      security_config
```
dasch: I would replace this with:

```ruby
{
  "security.protocol" => security_protocol,
  ...
}.compact
```

Nil values will be removed by `compact`.

README.md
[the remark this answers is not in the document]
dasch: It's actually something I use in production :/ it's nice for knowing how long a consumer group can be suspended for before it loses its offsets...

lib/racecar/consumer_set.rb
```diff
@@ -0,0 +1,90 @@
...
+      @consumers = []
+      @consumer_iterator = [].cycle
+    end
+
+    def subscribe
+      raise ArgumentError, "Subscriptions must not be empty when subscribing" if @config.subscriptions.empty?
+      @consumers = @config.subscriptions.map do |subscription|
+        consumer = Rdkafka::Config.new(rdkafka_config(subscription)).consumer
+        consumer.subscribe subscription.topic
+        consumer
```
dasch: Does an Rdkafka consumer not support multiple subscriptions?

dasch: Ah, is it because rdkafka-ruby doesn't support per-subscription config? That's supported by the protocol though – does librdkafka not support that?

JHK: It does, but not with different configuration options (like `auto.offset.reset`, which is quite handy to define in the handler for different topics).

lib/racecar/config.rb
JHK: Wasn't aware. Learning ruby here 😉 👍

lib/racecar/config.rb
JHK: Just convenience, because I found [a piece of documentation](https://github.com/zendesk/racecar#producing-messages-1) describing the config value being either string or symbol. Since rdkafka does not care I thought about rather being consistent and allowing both.

README.md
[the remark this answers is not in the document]
JHK: Okay, I've put it into my TODO list for going from alpha to beta. This one might be tricky to implement trough all the layers involved though.

lib/racecar/consumer_set.rb
dasch: :+1:

dasch: We'll see if we run into performance issues with many subscriptions then.

README.md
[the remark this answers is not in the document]
JHK: I found [this issue](https://github.com/edenhill/librdkafka/issues/594) being related to that topic. They once thought about implementing it per commit until this was removed from the protocol definition.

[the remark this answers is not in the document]
dasch: Hmm. It's still supported by ruby-kafka...

lib/racecar/config.rb
dasch: I think KingKonf should automatically cast the value to a Symbol in this case, though – is that not the case?

README.md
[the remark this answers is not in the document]
tjwp: Looks like it is part of the latest `OffsetCommit` request (version 4), but not part of `TxnOffsetCommit`. http://kafka.apache.org/protocol.html#The_Messages_OffsetCommit

lib/racecar/config.rb
dasch: Let me fix that in KingKonf.

dasch: Will have a fix ready Monday.

dasch: How does this look? https://github.com/dasch/king_konf/pull/14

thijsc:
Awesome that you're making this move. I'm a maintainer of `rdkafka-ruby` and happy to help out if anything comes up.

mensfeld:
@thijsc from my perspective the only thing that is missing in `rdkafka-ruby` is a batch support. It can be emulated like @JHK did but having in in the `rdkafka-ruby` would help a lot.

thijsc:
Took a quick look at d00dbc733680318b65ba2b5e28e36a7cfec81086, but not sure how this works exactly. What are you achieving with these batches?

lib/racecar/consumer_set.rb
thijsc: That's correct, there can be only one config per client. This is a limitation imposed by the C lib.

JHK:
@thijsc I tried to explain my use case in an issue on rdkafka-ruby: https://github.com/appsignal/rdkafka-ruby/issues/57

dasch:
@thijsc we've got a ruby-kafka + racecar + delivery boy Slack channel; do you want to join so that we can discuss things there?

https://ruby-kafka-slack.herokuapp.com/

mensfeld:
@dasch +karafka ;)

dasch:
Has that become the official channel for Karafka? :D

mensfeld:
@dasch for slack, why not ;) I think that having the ruby-kafka community in one place is better than in many ;)

thijsc:
Oh god another Slack :-). Will join!

JHK:
I've updated the remaining open items I had when creating this PR. For further development I've created the following issues on `rdkafka-ruby` to get on a similar level as before with `ruby-kafka`:
 * [Poll in batches #57](https://github.com/appsignal/rdkafka-ruby/issues/57)
 * [Ability to pause partitions #55](https://github.com/appsignal/rdkafka-ruby/issues/55)
 * [Instrumentation support #54 ](https://github.com/appsignal/rdkafka-ruby/issues/54)

lib/racecar/consumer_set.rb
```diff
@@ -18,24 +18,25 @@ def subscribe
...
-    rescue Rdkafka::RdkafkaError => e
-      raise unless e.is_partition_eof?
-      @logger.debug "No more messages on this partition."
-      @consumer_iterator.next
-      @messages.compact
+      while collect_messages_for_batch? do
+        msg = poll(timeout_ms)
+        break if msg.nil?
+        @messages << msg
+      end
```
dasch: Is this something we can push into rdkafka or librdkafka?

breunigs: I'm not sure I understood exactly what you're referring to, but I have the following suggestion:

librdkafka recently switched the default for `enable.partition.eof` to `false` https://github.com/edenhill/librdkafka/commit/a869fa19359fc285b19496c4321abe199322a736#diff-ab78e771fd78692a15ed791e0269a2db

This means there will be no more partition eof errors whatsoever and racecar would just have to deal with the `nil` case. For the transition period racecar should simply disable `enable.partition.eof` itself. Does that sound reasonable?

That being said, there's https://github.com/appsignal/rdkafka-ruby/issues/12 about the same issue. Considering it ought to be a simple wrapper around the C api of librdkafka, I'm not sure adding a high level `each`/`poll` makes sense there.

gaurish:
What is remaining here or blockers? Anything we could to push this along? As using native librdkafka would have immense benefits.

JHK:
### Features
I consider this PR is almost complete:
 * Pause support requires another release of `rdkafka-ruby`. Once there is one we wanted to bump the version here. See also https://github.com/appsignal/rdkafka-ruby/pull/78
 * Further instrumentation support: See https://github.com/appsignal/rdkafka-ruby/issues/54
 * There are conflicts :wink:

### Stability 
We are running this PR in production for some weeks now using Kafka 2.2. `rdkafka-ruby` hasn't been the source of any problems we've faced. I plan to publish a blogpost for our use case we had to work on this soon.

### Breaking Changes
This change requires a major release as it introduces some breaking changes. Those need to get documented carefully and sorted out with the core maintainers of this gem. See also https://github.com/zendesk/racecar/pull/97/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

### Documentation
The breaking changes require some re-write of the README. Those texts also need to get approved by the core maintainers of this gem. See also https://github.com/zendesk/racecar/pull/97/files#diff-04c6e90faac2675aa89e2176d2eec7d8

----

As we run this code on production already we consider it good enough for our use-case. Obviously we also would have a benefit when this code re-joins the main branch, but we would happily be willing to accept help to make this happen. @dasch did I conclude correct here or have I missed something from your point of view?

thijsc:
I just released rdkafka-ruby with all the latest additions. (pause was already in 0.5.0 by the way). I haven't made any progress on the instrumentation support. It's not something we have a need for ourselves, so haven't felt a lot of urgency.

If it's a bocker to get this pull shipped we should get it done though. I'll take a look at it this week.

mensfeld:
@thijsc I have work in progress with something similar for Waterdrop and Karafka, the instrumentation hooks would be really helpful.

JHK:
FYI: Here we have described our use-case that lead to this PR: https://tech.xing.com/asynchronous-state-replication-9e504aeac0de

dasch:
@JHK I am returning to work next week and will give this a review. Thanks for all your hard work!

lib/racecar/consumer_set.rb
```diff
@@ -0,0 +1,177 @@
...
+
+    def poll(timeout_ms)
+      maybe_select_next_consumer
+      retried ||= false
+      msg = current.poll(timeout_ms)
+    rescue Rdkafka::RdkafkaError => e
+      @logger.error "Error for topic subscription #{current_subscription}: #{e}"
+
+      case e.code
+      when :max_poll_exceeded, :"err_-147?" # Note: requires unreleased librdkafka version
```
dasch: Is this still the case?

lib/racecar/consumer_set.rb
```diff
@@ -0,0 +1,177 @@
...
+        retried = true
+        retry
+      else
+        raise
+      end
+    ensure
+      @last_poll_read_nil_message = true if msg.nil?
+    end
+
+    # XXX: messages are not guaranteed to be from the same partition
```
dasch: I think that could break existing consumers.

dasch:
Can you rebase and resolve the conflicts?

I think the best plan for rolling this out is to cut a `v1.0.0` release from the current master and create a stable branch, i.e. `v1-stable`. Then we merge this PR into `master` and cut `v2.0.0.alpha1`. 

@JHK will you be available for helping solve the many issues that will undoubtedly crop up? I unfortunately have very limited bandwidth.

lib/racecar/consumer_set.rb
breunigs: Oh, it certainly did. However, `ConsumerSet` is an entirely new class and we fixed this issue in `Racecar::Runner` in this commit: https://github.com/zendesk/racecar/pull/97/commits/c8902cff80cfe63892730277d4e2bf9372b04516#diff-2b2ca276543198a74d7942069ba0b48bR45

On `master` `Racecar::Runner#consumer` is public, but in this branch it's actually private, so it would break existing custom users already I think. Additionally the methods are named `each_batch` (master) and `batch_poll` (this branch).

I'm aware the behaviour change is not ideal, especially because it is potentially performance degrading (see linked commit message). If someone decides to use the `ConsumerSet#batch_poll` directly, they might be in for a surprise because it's not necessarily visible in testing. Naming the method something like `batch_poll_from_subscribed_partitions` would make it clearer, but it's not an exactly nice name. I'm fine with renaming this because it's not used often, but ideally this would be solved directly within librdkafka.

dasch: @thijsc would this be something that librdkafka could handle?

breunigs: I asked in `librdkafka`'s channel if there is a recommended approach to handle this in general (https://gitter.im/edenhill/librdkafka?at=5d4d71d21db76d0ffea12263). Let's see what they recommend, maybe I'm just missing some obvious batch poll method that supports this more or less out of the box.

breunigs:
+1 on the rollout/branch naming approach

Regarding availability: yes, we do have some time to support that. We don't have dedicated time for this, so I can offer best effort here, but I can't guarantee some fixed hours/week availability. Also most of the support would be from me, not @JHK.

dasch:
OK, I will probably start that work next week.

lib/racecar/consumer_set.rb
breunigs: from the chat:
> @breunigs what you can do is extract the per-partition queue, see queue_get_partition() and forward it to NULL and poll each partition queue explicitly.

Header for `queue_get_partition` is here: https://github.com/edenhill/librdkafka/blob/675e764b0fb013835c3ead74415bdb62b0aa7903/src/rdkafka.h#L2859-L2872

I took a brief look and this seems to allow a very fine grained setup of how message are consumed. Since there's a rebalance callback, old queues can probably be destroyed through that, but maybe `queue_get_partition` → `poll` → `queue_destroy` in the main loop is easier.
librdkafka also works on all assigned partitions semi-simultaneously, so we'd need something similar to switch queues if we stall on a single partition, or introduce a max time to ensure partitions are worked off more or less evenly. There'd also need to be a call to determine how many potential partitions a topic has, or at least how many we're currently subscribed to.

Apart from handling the details, exposing this function through rdkafka-ruby sounds reasonable, but reimplementing the polling/queue switching handling is probably on racecar side. 

Judging from gut feeling only, this seems effort and not a good trade-off. It took me a couple of iterations until I had all the bugs worked out on another feature which works with the lower-level APIs because it requires a very good understanding of Kafka and librdkafka. However, for us this non-optimal batching also incurs only a small performance penalty, so it was much easier to just start an additional process.

dasch:
Can you rebase and move the changelog entries under `v2.0.0`? Then I'll merge this and release an alpha.

Elyahou:
Is there a reason for this switch ?

dasch:
Yup – keeping ruby-kafka up to date with all the Kafka changes while avoiding regressions is too much work, and that's not even counting the missing features, such as background heartbeats, that we're still missing. By building upon rdkafka we can focus on ergonomic features rather than low-level stuff, and we should gain a lot of performance at the same time.

PR #111, opened 2019-03-10
Allow using Bundler 2

olleolleolle:
This PR allows installing the gem using latest bundler.


bquorning:
It’s a development dependency, so you can install with any version of Bundler. But I don’t know if there is any reason for having a Bundler version restriction at all. @dasch?

olleolleolle:
@bquorning The failure I got on Ruby 2.6.1 with Bundler 2 installed was an installation failure on `bundle install` in this repo.

Issue #114, opened 2019-03-20
Do not put current directory in LOAD_PATH


mainameiz:
Hello everybody!
First of all, thanks for making such a good project. It is a great addition to such a low-level ruby-kafka :)

---
My question...

I have found a strange issue using racecar + [bootsnap](https://github.com/Shopify/bootsnap).

As you may know Bootsnap recursively scans everything in `$LOAD_PATH` and caches all paths to speed up application boot process.

Also I have found that racecar adds current directory to $LOAD_PATH - https://github.com/zendesk/racecar/commit/0849ecc75fb22649e640d26e337d0be53e4d1dd5.

This causes bootsnap to scan every directory in my rails project :) including shared storage linked to `public/uploads`.

So my question: is it necessary to put current directory into `$LOAD_PATH`?

mainameiz:
I also make a PR https://github.com/zendesk/racecar/pull/115

dasch:
It's done in order for `--require` to work, I believe.

mainameiz:
Oh, I understand. May be this can be done without loading current directory to `$LOAD_PATH`?

E.g.

```ruby
        opts.on("-r", "--require STRING", "Require a library before starting the consumer") do |lib|
          begin
            require lib
          rescue LoadError => _exception
            require File.join(Dir.pwd, lib)
          end
        end
```

What do you think?

dasch:
I think that will subtly change how Ruby loads files, which may cause files to be loaded twice. The string passed to `require` is used to check if a file has already been loaded.

dasch:
Does Rails not also put the current dir in the load path? Or is it just `lib/`?

dasch:
I think one way to deal with this would be to only add to the load pass if `--require` has been passed.

mainameiz:
> Does Rails not also put the current dir in the load path? Or is it just lib/?

Yes, just `/lib` is placed in $load_path, but not root directory.

dasch:
It would be a breaking change to change from current dir to `./lib/`, so I don't think that's an option – but if the modification of the load path can be made conditional on one or more `--require` args being passed then all can be happy?

mainameiz:
> I think one way to deal with this would be to only add to the load pass if --require has been passed.

Not quite understand. Do you mean this?

```ruby
        opts.on("-r", "--require STRING", "Require a library before starting the consumer") do |lib|
          begin
            require lib
          rescue LoadError => _exception
[...]

PR #115, opened 2019-03-20
Do not put current directory into $LOAD_PATH

mainameiz:


mainameiz:
https://github.com/zendesk/racecar/issues/114#issuecomment-478959737

PR #117, opened 2019-04-02
Add current dir to LOAD_PATH only when --require option is used

mainameiz:
Related issue https://github.com/zendesk/racecar/issues/114

Suggested solution - https://github.com/zendesk/racecar/issues/114#issuecomment-478960548


dasch:
Thanks!

PR #120, opened 2019-06-13
support for ssl_verify_hostname in the configuration

nadavk72:
ruby-kafka introduce new configuration property that solve a critical issue when connecting to Heroku with SSL: [SSL verification breaks Heroku Kafka #734](https://github.com/zendesk/ruby-kafka/issues/734)
Without this fix I wasn't able to connect to heroku's Kafka with racecar...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@
 /spec/reports/
 /tmp/
 /vendor/bundle/
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /spec/reports/
 /tmp/
 /vendor/bundle/
+.idea/
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -26,7 +26,7 @@ module Racecar
       RailsConfigFileLoader.load!
 
       if File.exist?("config/racecar.rb")
-        require "config/racecar"
+        require "racecar/config"
       end
 
       # Find the consumer class by name.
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 /spec/reports/
 /tmp/
 /vendor/bundle/
+.idea/
```
dasch: Can you remove this change?

lib/racecar/cli.rb
```diff
@@ -26,7 +26,7 @@ def run
       RailsConfigFileLoader.load!
 
       if File.exist?("config/racecar.rb")
-        require "config/racecar"
+        require "./config/racecar"
```
dasch: Is this change relevant?

lib/racecar/version.rb
```diff
@@ -1,3 +1,3 @@
 module Racecar
-  VERSION = "0.5.0.beta2"
+  VERSION = "0.5.4.beta2"
```
dasch: Can you remove this change?

.gitignore
nadavk72: done

lib/racecar/cli.rb
nadavk72: without this change I wasn't able to run my forked version. it keeps searching the config file under the original gem...

lib/racecar/version.rb
nadavk72: done

sparrovv:
hey 👋 I'm not sure why, but those changes are not available in `0.5.0.beta2`, could we bump it?

Issue #119, opened 2019-06-05
False-Positive when detecting rails application


Bartuz:
TL;DR
`require 'rails'` returns true for non-rails project with "activesuppport" or "railties"
----

Detection if the app is a rails app happens [here](https://github.com/zendesk/racecar/blob/master/lib/racecar/rails_config_file_loader.rb#L6-L9)

which produces false positive when app uses activesupport. Rails object is then defined, but it has no "logger" object so [these lines](https://github.com/zendesk/racecar/blob/master/lib/racecar/rails_config_file_loader.rb#L23-L28) raise an exception (undefined method on nil object :) )

It was detected in a closed-source project, I can share part of my Gemfile.lock

```
$ cat Gemfile.lock G rail
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.2)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.3)
    rails-dom-testing (2.0.3)
    rails-html-sanitizer (1.0.4)
    railties (5.1.6)
      railties (>= 4.2.7, < 5.3.0)
```

and

```
$ cat Gemfile.lock G active
      activesupport (= 5.1.6)
      activesupport (= 5.1.6)
    activemodel (5.1.6)
      activesupport (= 5.1.6)
    activerecord (5.1.6)
      activemodel (= 5.1.6)
      activesupport (= 5.1.6)
    activesupport (5.1.6)
      activesupport (>= 4.2.0)
      activesupport (= 5.1.6)
      activerecord (>= 4.2.7, < 5.3.0)
  activerecord (~> 5.1)
  activesupport (~> 5.1)
```

Bartuz:
Our temporary fix which apparently does not break anything


```
#kafka_setup.rb

Racecar::RailsConfigFileLoader.class_eval do
  def self.load!
    $stderr.puts "=> Skipping rails detection as monkeypatch in kafka_setup.rb..."
  end
end
```

and include this file in `bundle exec racecar --require kafka_setup.rb`

dasch:
Can you figure out a better way of detecting Rails?

Bartuz:
In our case (our case = no rails but activesupport/railties) script fails on (Rails.logger.extend)[https://github.com/zendesk/racecar/blob/master/lib/racecar/rails_config_file_loader.rb#L27] so kind of ugly hot-fix would be changing
```
   begin
        require "rails"
[...]

PR #121, opened 2019-07-17
Implement a simple test and move this into the begin rescue block

simonireilly:
Fix for: #119 

### Context

We don't want to use any of rails if it is missing so I have just hoisted this if into the begin -> rescue.

I am sure there is some reason that this isn't a great fix, but I'd like to get the conversation started on this one as we need a fix :+1:

### DIFF
--- a/lib/racecar/rails_config_file_loader.rb
+++ b/lib/racecar/rails_config_file_loader.rb
@@ -5,27 +5,27 @@ module Racecar
 
       begin
         require "rails"
+
+        if defined?(Rails)
+          $stderr.puts "=> Detected Rails, booting application..."
+
+          require "./config/environment"
+
+          if (Rails.root + config_file).readable?
+            Racecar.config.load_file(config_file, Rails.env)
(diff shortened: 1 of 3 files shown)


lib/racecar/rails_config_file_loader.rb
```diff
@@ -5,27 +5,27 @@ def self.load!
...
+          # Unicorn does.
+          if Rails.env.development? && defined?(ActiveSupport::Logger)
+            console = ActiveSupport::Logger.new($stdout)
+            console.formatter = Rails.logger.formatter
+            console.level = Rails.logger.level
+            Rails.logger.extend(ActiveSupport::Logger.broadcast(console))
+          end
         end
+      rescue LoadError
+        # Not a Rails application.
```
mzishtiaq: If it's not a rails application then perhaps we should output something like this:
```
 $stderr.puts "=> Detected Ruby, booting application..."
```
Any thoughts regarding that?

PR #123, opened 2019-09-03
Make `produce` backwards compatible with v1

dasch:


dasch:
Yeah, but this one seems unnecessary.

PR #125, opened 2019-09-03
Pass `create_time` as the message timestamp

dasch:


dasch:
@JHK @breunigs this adds more backwards compatibility.

Issue #122, opened 2019-09-03
Exception: Unsupported compression type


dasch:
Got this when restarting Kafka in local dev:

```
Crashed: Rdkafka::RdkafkaError: Unsupported compression type (unsupported_compression_type)
/bundle/gems/rdkafka-0.6.0/lib/rdkafka/consumer.rb:355:in `poll'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/consumer_set.rb:16:in `poll'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/runner.rb:69:in `block (2 levels) in run'
/bundle/gems/activesupport-5.1.6.2/lib/active_support/notifications.rb:168:in `instrument'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/runner.rb:61:in `block in run'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/runner.rb:58:in `loop'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/runner.rb:58:in `run'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar.rb:50:in `run'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/cli.rb:63:in `run'
/bundle/gems/racecar-2.0.0.alpha1/lib/racecar/cli.rb:10:in `main'
/bundle/gems/racecar-2.0.0.alpha1/exe/racecar:7:in `<top (required)>'
bin/racecar:17:in `load'
bin/racecar:17:in `<main>'
```

/cc @JHK @breunigs

JHK:
Good one, we set it to `lz4` everywhere. I'll have a look

JHK:
I wonder how compression comes into play on the consumer side. According to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md the compression parameters can only be set for producers.

dasch:
🤷‍♂ 

I'm not even sure why this happened – it was during broker shutdown. Could be an incorrect error message.

breunigs:
I’m on mobile right now, but generally speaking printing the error code might be helpful. Depending on the versions in use there might be a disagreement on what an error code means, see e.g. https://github.com/edenhill/librdkafka/issues/2245

breunigs:
I don't know how to reproduce this.

[...]

PR #129, opened 2019-09-11
retry transport failures instead of crashing (can happen during broker shutdown)

breunigs:
This has been discussed in #122, although that bug is about another issue. 

In my local tests an exponential back off was not needed, since this error is apparently only triggered once when the connection is torn down.


dasch:
Thanks!

PR #130, opened 2019-09-18
Wrap the rdkafka message

dasch:
Provides backwards compatibility and avoids leaking internals.


mensfeld:
Is it worth it? It basically doubles the amount of objects created when working with racecar.

dasch:
I think it's worth it to avoid bleeding implementation details and maintain backwards compatibility.

PR #131, opened 2019-10-02
retry poll failures with exponential backoff (see #122)

breunigs:
As per https://github.com/zendesk/racecar/issues/122#issuecomment-530777250 , let's keep to the existing promises. 

It doesn't fix the specific issue mentioned in #122, though.

### DIFF
--- a/lib/racecar/consumer_set.rb
+++ b/lib/racecar/consumer_set.rb
@@ -14,11 +14,15 @@ module Racecar
 
     def poll(timeout_ms)
       maybe_select_next_consumer
+      started_at ||= Time.now
       try ||= 0
-      msg = current.poll(timeout_ms)
+
+      remain = remaining_time_ms(timeout_ms, started_at)
+      msg = remain <= 0 ? nil : current.poll(remain)
     rescue Rdkafka::RdkafkaError => e
+      wait_before_retry_ms = 100 * (2**try) # 100ms, 200ms, 400ms, …
       try += 1
-      raise if try >= MAX_POLL_TRIES
+      raise if try >= MAX_POLL_TRIES || remain <= wait_before_retry_ms
 
       @logger.error "(try #{try} of #{MAX_POLL_TRIES}): Error for topic subscription #{current_subscription}: #{e}"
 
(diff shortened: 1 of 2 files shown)


dasch:
Awesome!

I do think, however, that we need to respect `timeout_ms`. How about stopping retries once that time interval has passed? Otherwise we could be looking at very long latencies for the call.

lib/racecar/consumer_set.rb
```diff
@@ -12,21 +14,21 @@ def initialize(config, logger)
 
     def poll(timeout_ms)
       maybe_select_next_consumer
```
dasch: I’m thinking we set a deadline here and check it before retrying.

breunigs:
You bring up a valid point, but I'm wondering about decent default values. If the racecar runner is used, then the value here is `config.max_wait_time` which is:
```
    desc "How long to allow the Kafka brokers to wait before returning messages"
    float :max_wait_time, default: 1 # seconds
```
This value is also used for batch polling, so the 1 second seems a decent trade-off between ability to batch and lag. To avoid exceeding the limit too much, I'll amend the code to pass the remaining time budget instead of the full budget for every poll. 

With the current settings there'll be at most 1 retry, which seems undesirable. Adjusting the calculation to e.g.
```ruby
sleep (2**(try-1))/10
```
will give sleeps of 100ms, 200ms, 400ms, 800ms, …, 51.2s. Assuming a failing poll doesn't take too long, there'd be at least 4 tries, which seems reasonable. Well, 5, if we decide the additional retry is worth more than exceeding the time limit. I'd err on early failure here, though.

dasch:
I think that's a great approach :+1:

breunigs:
Note that the code right now has a race condition:
1. has enough remaining time to attempt retry
2. something in `rescue` block is slow (e.g. logger being throttled)
3. next try re-evaluates remaining time and simply returns `nil` 

I'll see if I can find a decent fix

dasch:
@breunigs are you happy with the current version? Ready to merge?

breunigs:
Yes, I was about to write a "ready for review" comment to clean up any misunderstandings, because I consider this good to go now.

dasch:
Thanks!

PR #134, opened 2019-10-16
fix Consumer#produce method interface in docs

JanaVPetrova:
seems like it was changed in https://github.com/zendesk/racecar/pull/123
but the docs was not updated


dasch:
Thanks!

PR #137, opened 2019-12-09
Add Option To Not Boot Rails When Starting Consumer

apeacock1991:
In the case of simple consumers, there would be no need to boot Rails to process the messages, as you wouldn't necessarily need Rails. This is particularly advantageous in larger Rails applications, which would use significantly more resources than a simple consumer running vanilla Ruby.

I benchmarked the consumer we use at the company I work for - both with and without starting Rails. With Rails it uses 282M whereas without it uses 23M, just under a 92% saving in terms of memory alone.

### DIFF
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -23,7 +23,7 @@ module Racecar
     def run
       $stderr.puts "=> Starting Racecar consumer #{consumer_name}..."
 
-      RailsConfigFileLoader.load! unless config.disable_rails?
+      RailsConfigFileLoader.load! unless config.without_rails?
 
       if File.exist?("config/racecar.rb")
         require "./config/racecar"
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -23,7 +23,7 @@ module Racecar
     def run
       $stderr.puts "=> Starting Racecar consumer #{consumer_name}..."
 
-      RailsConfigFileLoader.load!
+      RailsConfigFileLoader.load! unless config.norails
 
       if File.exist?("config/racecar.rb")
         require "./config/racecar"
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -23,7 +23,7 @@ module Racecar
     def run
       $stderr.puts "=> Starting Racecar consumer #{consumer_name}..."
 
-      RailsConfigFileLoader.load! unless config.norails?
+      RailsConfigFileLoader.load! unless config.disable_rails?
 
       if File.exist?("config/racecar.rb")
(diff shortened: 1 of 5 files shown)


lib/racecar/config.rb
```diff
@@ -136,6 +136,9 @@ class Config < KingKonf::Config
     desc "Whether to check the server certificate is valid for the hostname"
     boolean :ssl_verify_hostname, default: true
 
+    desc "Whether to boot Rails when starting the consumer"
+    boolean :norails, default: false
```
dasch:
```suggestion
    boolean :no_rails, default: false
```

lib/racecar/cli.rb
```diff
@@ -23,7 +23,7 @@ def config
     def run
       $stderr.puts "=> Starting Racecar consumer #{consumer_name}..."
 
-      RailsConfigFileLoader.load!
+      RailsConfigFileLoader.load! unless config.norails?
```
dasch:
```suggestion
      RailsConfigFileLoader.load! unless config.no_rails?
```

spec/cli_spec.rb
```diff
@@ -1,7 +1,30 @@
...
+  # has to be reset between examples otherwise it caches across examples and causes unexpected behavior
+  before do
+    Racecar.config = nil
+  end
+
   it "fails if no consumer class is specified" do
     expect { Racecar::Cli.main([]) }.to raise_exception(Racecar::Error, "no consumer specified")
   end
+
+  it "doesn't start Rails if --norrails option is specified" do
```
dasch:
```suggestion
  it "doesn't start Rails if --no-rails option is specified" do
```

spec/cli_spec.rb
```diff
@@ -1,7 +1,30 @@
...
+    args = ["--require", "./examples/cat_consumer.rb", "CatConsumer", "--norails"]
```
dasch:
```suggestion
    args = ["--require", "./examples/cat_consumer.rb", "CatConsumer", "--no-rails"]
```

spec/cli_spec.rb
```diff
@@ -1,7 +1,30 @@
...
+
+    allow(Racecar).to receive(:run)
+    expect(Racecar::RailsConfigFileLoader).not_to receive(:load!)
+
+    Racecar::Cli.main(args)
+  end
+
+  it "starts Rails if the --norails option is omitted" do
```
dasch:
```suggestion
  it "starts Rails if the --no-rails option is omitted" do
```

apeacock1991:
@dasch Thanks for the review! Amended the parameter to `--disable-rails`, is it all good now?

PR #142, opened 2020-01-27
Datadog instrumentation

angelim:
Moves over `ruby-kafka` instrumentation to Racecar. There are still a lot of metrics missing from the original. I could only address what's currently exposed by `rdkafka`.
It's mostly copy/paste with a few tweaks.

Some notes for review:
* `racecar.consumer.loop.duration` is really noisy. I haven't tested how it was before, but worth taking a look.
* I'm not sure whether we should use `buffer_size:  @delivery_handles.size`. This is the local buffer and not the one from librdkafka. Hoping for some ideas.
* `join_group` and `leave_group` instrumentation might be covering more ground than what they used to in ruby-kafka.

We also need to organize this properly. Splitting subscribers looks like a natural next step.

### DIFF
--- a/lib/racecar.rb
+++ b/lib/racecar.rb
@@ -1,5 +1,6 @@
 require "logger"
 
+require "racecar/instrumenter"
 require "racecar/null_instrumenter"
 require "racecar/consumer"
 require "racecar/consumer_set"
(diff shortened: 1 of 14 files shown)


dasch:
Awesome! I’ll review this next week!

PR #145, opened 2020-02-06
fix: Call error handler with exception

cmschuetz:
This ensures that the global error handler is called when a runner is configured with a pause timeout.  I've also updated the README to remove the `info` argument as I don't see `error_handler` called with it anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,12 +455,8 @@ If, on the other hand, the exception was cause by a temporary network or databas
 In addition to retrying the processing of messages, Racecar also allows defining an _error handler_ callback that is invoked whenever an exception is raised by your `#process` method. This allows you to track and report errors to a monitoring system:
 
 ```ruby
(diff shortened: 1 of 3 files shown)


dasch:
I actually think it's a regression that we don't get the `info` parameter – it's critical to identifying the specific message that caused the problem, which might be crucial for debugging (e.g. malformed messages).

Could you amend you PR so that we correctly pass in this parameter? I think we need to wrap `processor.process` with a `rescue` that calls the error handler before re-raising.

cmschuetz:
@dasch I agree, it's really valuable.  I've updated this PR with your suggestions - I couldn't wrap the entire `process` / `process_batch` methods with a rescue since `with_pause` will pick it up first.  Is this what you had in mind?

dasch:
It is! Thanks!

PR #147, opened 2020-02-20
Fix missing constant error

jonmagic:
## Why?

Because dogstatsd-ruby moved some constants around. I was getting these errors when running specs:

> uninitialized constant Datadog::Statsd::Connection::DEFAULT_HOST

## How?

Apply the suggestion from https://github.com/zendesk/ruby-kafka/pull/810#issuecomment-589415255

Note: Similar to the change I made in https://github.com/zendesk/ruby-kafka/pull/810 but only had to go back one minor version instead of 2.

### DIFF
--- a/lib/racecar/datadog.rb
+++ b/lib/racecar/datadog.rb
@@ -26,7 +26,7 @@ module Racecar
       end
 
       def host
-        @host ||= default_host
+        @host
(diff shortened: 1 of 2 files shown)


dasch:
Thanks!

PR #148, opened 2020-02-20
Add time_lag metric to process_batch instrumentation_payload

jonmagic:
## Why?

As explained in https://github.com/zendesk/ruby-kafka/pull/811 we're excited to try out racecar 2 beta but need a metric for lag in the process_batch instrumentation. It doesn't look like offset_lag will be possible initially but the time_lag metric available in the process_message metrics would be useful to us if added to the process_batch metrics

## How?

Adds time_lag to process_batch metrics.


dasch:
Thanks!

Issue #143, opened 2020-01-29
Using racecar CLI swallows errors raised from required libs


thetizzo:
Hi 👋,

I'm using Racecar 1.0.1 to build a consumer library and I'm having an issue where running racecar from the command line seems to silently swallow any error raise by a required library during boot.

I'm running it with this command, as per the README:

```bash
$ bundle exec racecar --require <my_library> MyConsumer
```

The error being raised by my library happens when the library detects a missing required environment variable during the start up process but Racecar output nothing to logs or console and just exits the process.  

This seems very similar to issue #106 but that seemed to be closed with no resolution for this so I thought I would reopen it.

My proposal to fix this would be either of the following:

1: In the `exe/racecar` script file, just add a log message that outputs the error that was raised to cause the rescue.  Something like this:

```ruby
begin
  Racecar::Cli.main(ARGV)
rescue SignalException => e
  # ...
rescue => e
  $stderr.puts "New helpful message: #{e.message}" # New message here
  exit(1)
else
  exit(0)
end
``` 

2: In `Racecar::Cli#build_parser`, change the block in the options parser to have a rescue for the require of the new library in case it doesn't load for some reason.  Something like this:
```ruby
opts.on("-r", "--require STRING", "Require a library before starting the consumer") do |lib|
  $LOAD_PATH.unshift(Dir.pwd) unless load_path_modified
  load_path_modified = true
  begin
    require lib
  rescue => e
    $stderr.puts "New helpful message: #{e.message}"
    exit
  end
end
```

I like the second option because it's more specifically solving this issue of a library failing to load so we can make the log message there more helpful.  

Let me know what you think and I can provide a PR for `v1-stable` or `master` or both with the fix.

dasch:
I like option 2 as well – can you provide a fix for both branches?

thetizzo:
Sound good!  I will have PRs for this incoming soon.

PR #149, opened 2020-02-21
[v1-stable] Added error handling for required libraries

thetizzo:
If a required library failed to load then the failure would be silently swallowed.  The error handling added here now sends the error message to STDERR.

This PR addresses #143 for the `v1-stable` branch.


dasch:
Thanks!

PR #150, opened 2020-02-21
[master] Added error handling for required libraries

thetizzo:
If a required library failed to load then the failure would be silently swallowed.  The error handling added here now sends the error message to STDERR.

This PR addresses #143 for the `master` branch.

### DIFF
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -102,7 +102,12 @@ module Racecar
         opts.on("-r", "--require STRING", "Require a library before starting the consumer") do |lib|
           $LOAD_PATH.unshift(Dir.pwd) unless load_path_modified
           load_path_modified = true
-          require lib
+          begin
+            require lib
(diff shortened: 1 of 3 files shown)


lib/racecar/cli.rb
```diff
@@ -102,7 +102,12 @@ def build_parser
         opts.on("-r", "--require STRING", "Require a library before starting the consumer") do |lib|
           $LOAD_PATH.unshift(Dir.pwd) unless load_path_modified
           load_path_modified = true
-          require lib
+          begin
+            require lib
+          rescue => e
+            $stderr.puts "=> Required library failed to load: #{e.message}"
```
dasch: Can you include the name of the library in the message?

thetizzo: Updated this to reference the library name instead of saying "Required library."

dasch:
Thanks!

PR #151, opened 2020-02-25
SSL Configuration

ellimist:
#### Description
Add a new configuration options: `ssl_certificate_location`, `ssl_key_location` and `ssl_key_password` that will get passed down to rdkafka in order to configure the certificate location.

In order to configure properly the authentication, we need to specify a certificate path too, which we currently do not: https://github.com/edenhill/librdkafka/wiki/Using-SSL-with-librdkafka#configure-librdkafka-client

Also, per the librdkafka documentation linked above ☝️ from the client perspective, we should be sending `ssl.key.location` and `ssl.key.password` too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,9 @@ The memory usage limit is roughly estimated as `max_bytes * min_message_queue_si
 * `ssl_crl_location` – Path to CRL for verifying broker's certificate validity
 * `ssl_keystore_location` – Path to client's keystore (PKCS#12) used for authentication
 * `ssl_keystore_password` – Client's keystore (PKCS#12) password
+* `ssl_certificate_location` – Path to the certificate
+* `ssl_key_location` – Path to client's certificate used for authentication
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -286,8 +286,9 @@ The memory usage limit is roughly estimated as `max_bytes * min_message_queue_si
 * `security_protocol` – Protocol used to communicate with brokers (`:ssl`)
 * `ssl_ca_location` – File or directory path to CA certificate(s) for verifying the broker's key
 * `ssl_crl_location` – Path to CRL for verifying broker's certificate validity
-* `ssl_keystore_location` – Path to client's keystore (PKCS#12) used for authentication
-* `ssl_keystore_password` – Client's keystore (PKCS#12) password
```
dasch: We shouldn't remove existing options unless you're sure they don't work at all – others will rely on them.

ellimist: 👍 added the new options without removing the old ones instead.

PR #157, opened 2020-02-26
Fix `time_lag` computation

dasch:


lib/racecar/datadog.rb
```diff
@@ -89,7 +89,7 @@ class ConsumerSubscriber < StatsdSubscriber
       def process_message(event)
         offset = event.payload.fetch(:offset)
         create_time = event.payload.fetch(:create_time)
-        time_lag = create_time && ((Time.now - create_time) * 1000).to_i
+        time_lag = create_time && ((Time.now - (create_time * 1000)).to_i
```
jonmagic: If `create_time` is a `Time` object and `Time.now` is a `Time` object subtracting `create_time` from `Time.now` should give you seconds. Then you can multiple the result by 1000 to get milliseconds and then call `to_i` to remove the fractional milliseconds.

```ruby
[1] pry(main)> a = Time.now
=> 2020-02-27 11:36:50 -0800
[2] pry(main)> b = Time.now
=> 2020-02-27 11:36:54 -0800
[3] pry(main)> b - a
=> 3.840914
[4] pry(main)> (b - a) * 1000
=> 3840.914
[5] pry(main)> ((b - a) * 1000).to_i
=> 3840
```

So I think it was technically correct before?

jonmagic: I should also note we're using the current implementation in v2beta4 in staging and production right now and getting the values we would expect from the time_lag metric.

dasch: :+1:

PR #162, opened 2020-04-17
Fix example for processing messages in batches

vk26:
Hi. Current example in docs for processing messages in batches is not worked. I fixed it.


vk26:
I saw It isn't actual for v2 (master). Close it.

PR #163, opened 2020-04-21
Rename message.max.bytes configuration

bastien:
According to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md this setting should be called `message.max.bytes` instead of `fetch.message.max.bytes` which explains why if you set `fetch.max.bytes` to something smaller than the default `message.max.bytes` you see this error:

```
`fetch.max.bytes` must be >= `message.max.bytes` (Rdkafka::Config::ClientCreationError)
```


dasch:
Thanks!

PR #164, opened 2020-04-22
Require ruby-kafka v1.0 or higher

tjwp:
This change is against the v1-stable branch to create a racecar version that can be used with ruby-kafka v1.0.

I've also updated the CHANGELOG with recent changes.


dasch:
Thanks!

PR #166, opened 2020-04-30
Bring back `partition_key`

dasch:
This should ensure backwards compatibility with Racecar v1.

PR #168, opened 2020-04-30
Upgrade rdkafka to v0.7.0

dasch:


puneetloya:
is there a release date planned for 2.0.0?

dasch:
Once #166 has been merged, v2 should be API backwards compatible with v1. The next challenge could be how to do the actual v1 -> v2 upgrade for existing consumer groups, since the client-specific protocol has changed. Not sure how best to do that.

PR #176, opened 2020-07-29
add instrumentation at the start of main_loop

breunigs:
Akin to the other instrumentations `start_process_message`/`start_process_batch`, allow to subscribe to start of main loop. My use case is to more easily connect a "request oriented" logging framework in use at my org, i.e. I have an actual use case for this.

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -61,8 +61,10 @@ module Racecar
       # Main loop
       loop do
         break if @stop_requested
-        resume_paused_partitions
+        @instrumenter.instrument("start_main_loop", instrumentation_payload)
         @instrumenter.instrument("main_loop", instrumentation_payload) do
+          resume_paused_partitions
+
           case process_method
           when :batch then
             msg_per_part = consumer.batch_poll(config.max_wait_time).group_by(&:partition)
--- a/spec/runner_spec.rb
+++ b/spec/runner_spec.rb
@@ -415,6 +415,14 @@ RSpec.describe Racecar::Runner do
         runner.run
       end
 
+      it 'instruments the start of main_loop' do
+        expect(instrumenter).
+          to receive(:instrument).
+          with("start_main_loop", hash_including(loop_instrumentation))
+
+        runner.run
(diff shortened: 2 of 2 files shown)


lib/racecar/runner.rb
```diff
@@ -61,8 +61,10 @@ def run
       # Main loop
       loop do
         break if @stop_requested
-        resume_paused_partitions
+        @instrumenter.instrument("start_main_loop", instrumentation_payload)
```
tjwp, line 64: Do you need a separate event for this or could you create a subscriber that performs whatever logic is required at the start of the existing `main_loop` instrumentation:

```ruby
class StartMainLoopSubscriber
  def start(name, id, payload)
     log_request(...)
  end

  # Required to use the evented interface
  def finish(_name, _id, _payload); end
end

ActiveSupport::Notifications.subscribe("main_loop.racecar", StartMainLoopSubscriber.new)
```

breunigs: Indeed! I wasn't aware of this possibility and couldn't find it when I searched for it. Do you have any documentation links handy? Looking at the docs for the current Rails release (https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) I don't see any mention of it, either. It even unhelpfully suggests above method, but says the class should implement `call`. I'm wondering if this is just documented badly or sort of a private interface.

tjwp: I'm pretty new to `ActiveSupport::Notifications` but had the same experience trying to find documentation on this. I'm leaning towards this just being poorly documented. I first learned about this interface from looking at how Datadog's library implements tracing based on subscribers: https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/active_support/notifications
Comments in that code indicate that ActiveSupport 3.x used `call`, but ActiveSupport 4+ uses the start and finish methods. I believe that the object that gets subscribed is eventually passed through to here as the listener: https://github.com/rails/rails/blob/master/activesupport/lib/active_support/notifications/fanout.rb#L91

PR #177, opened 2020-07-31
fix edge case where no retries were done for batch_poll

breunigs:
The edge case looks something like this:
1. `batch_poll` iterates just fine until a very short amount of time is left, say 1ms
2. It will trigger another `poll`, which can of course fail
3. there is no time left to do any retry, thus immediately it will immediately re-raise and let the error bubble

Since the code was tough to follow already, I decided to refactor it instead of hacking the `MIN_POLL_TRIES` feature in. I adjusted the signature of `batch_poll`, even though that is not strictly needed. It just seemed weird to require `max_wait_time`/`timeout_ms` to be passed in, while the `max_messages` was "hard"coded to `@config.fetch_messages`. It's backwards compatible.

There's also another change, since rdkafka now always gets passed the `max_wait_time` instead of the `remain_ms`. This is to give it enough time for a real retry, e.g. when only 1ms remains and there were connection hiccups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * [Bugfix] Close RdKafka consumer in ConsumerSet#reset_current_consumer to prevent memory leak (#196)
+* [Bugfix] `poll`/`batch_poll` would not retry in edge cases and raise immediately. They still honor the `max_wait_time` setting, but might return no messages instead and only retry on their next call. ([#177](https://github.com/zendesk/racecar/pull/177))
 
 ## racecar v2.1.0
 
--- a/lib/racecar/consumer_set.rb
+++ b/lib/racecar/consumer_set.rb
@@ -126,12 +126,12 @@ module Racecar
 
       poll_current_consumer(max_wait_time_ms)
     rescue Rdkafka::RdkafkaError => e
-      @logger.error "(try #{try}): Error for topic subscription #{current_subscription}: #{e}"
-
       remain_ms = remaining_time_ms(max_wait_time_ms, started_at)
       wait_ms = 100 * (2**try) # 100ms, 200ms, 400ms, …
 
       try += 1
+      @logger.error "(try #{try}): Error for topic subscription #{current_subscription}: #{e}"
+
       raise if try >= MIN_POLL_TRIES && wait_ms >= remain_ms
       raise if try >= MAX_POLL_TRIES
 
--- a/lib/racecar/consumer_set.rb
+++ b/lib/racecar/consumer_set.rb
@@ -1,5 +1,6 @@
 module Racecar
   class ConsumerSet
+    MIN_POLL_TRIES = 2
     MAX_POLL_TRIES = 10
 
     def initialize(config, logger, instrumenter = NullInstrumenter)
@@ -13,46 +14,34 @@ module Racecar
       @last_poll_read_nil_message = false
     end
 
-    def poll(timeout_ms)
-      maybe_select_next_consumer
-      started_at ||= Time.now
-      try ||= 0
-      remain ||= timeout_ms
-
-      msg = remain <= 0 ? nil : current.poll(remain)
-    rescue Rdkafka::RdkafkaError => e
-      wait_before_retry_ms = 100 * (2**try) # 100ms, 200ms, 400ms, …
-      try += 1
-      raise if try >= MAX_POLL_TRIES || remain <= wait_before_retry_ms
-
-      @logger.error "(try #{try}): Error for topic subscription #{current_subscription}: #{e}"
-
-      case e.code
-      when :max_poll_exceeded, :transport # -147, -195
-        reset_current_consumer
-      end
-
-      remain = remaining_time_ms(timeout_ms, started_at)
-      raise if remain <= wait_before_retry_ms
-
-      sleep wait_before_retry_ms/1000.0
-      retry
-    ensure
-      @last_poll_read_nil_message = true if msg.nil?
+    def poll(max_wait_time_ms = @config.max_wait_time)
+      batch_poll(max_wait_time_ms, 1).first
     end
 
-    # XXX: messages are not guaranteed to be from the same partition
-    def batch_poll(timeout_ms)
-      @batch_started_at = Time.now
-      @messages = []
-      while collect_messages_for_batch? do
-        remain = remaining_time_ms(timeout_ms, @batch_started_at)
-        break if remain <= 0
-        msg = poll(remain)
+    # batch_poll collects messages until any of the following occurs:
+    # - max_wait_time_ms time has passed
+    # - max_messages have been collected
+    # - a nil message was polled (end of topic, Kafka stalled, etc.)
+    #
+    # The messages are from a single topic, but potentially from more than one partition.
+    #
+    # Any errors during polling are retried in an exponential backoff fashion, roughly
+    # honoring the given time limit. It will try at least MIN_POLL_TRIES times,
+    # regardless of the given time limit.
+    def batch_poll(max_wait_time_ms = @config.max_wait_time, max_messages = @config.fetch_messages)
+      started_at = Time.now
+      remain_ms = max_wait_time_ms
+      maybe_select_next_consumer
+      messages = []
+
+      while remain_ms > 0 && messages.size < max_messages
+        msg = poll_with_retries(remain_ms)
         break if msg.nil?
-        @messages << msg
+        messages << msg
+        remain_ms = remaining_time_ms(max_wait_time_ms, started_at)
       end
-      @messages
+
+      messages
     end
 
     def store_offset(message)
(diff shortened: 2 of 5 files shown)


dasch:
This changes the behavior in a crucial way, though – before, the user could set an upper bound on the _total_ time spent in a blocking state. That's why `remain` was there.

breunigs:
This is true. There's these constraints or guarantees we'd ideally provide:
- honor time limit
- retry ephemeral errors
- eventually raise on non-recoverable errors
- maximize batching
- don't break the interface
- don't push the issue onto the user by providing more config options

Looking at librdkafka's code that eventually gets called on poll, we can see it sets a timeout error if there was no result (!= no message) within the specified time limit: https://github.com/edenhill/librdkafka/blob/329c97fe0f408a2f91471fa3d835861b2025d6ea/src/rdkafka.c#L2860-L2866 . Reasonable options to deal with this at a low level are a lower bound on the time limit, retry or ignore.

In most cases it provides results within 1ms, measured from Racecar's main_loop. For the discussion I'm going to assume a sensible lower bound is 100ms and max_time is 250ms. If we value the max time more, we can only poll if at least 100ms are left since we can't guarantee librdkafka will not actually use the time given to it. This approach is at odds with maximizing batching, since we'd likely get another 100 polls in the allotted time.

Assuming the conditions are right, we might try a poll with 1ms time left. If it fails, we cannot retry any more if we honor the time limit. This is obviously at odds with not raising on ephemeral errors.

Ignoring carries the risk of never raising, even if the issue persists.

Without a lower bound, we're likely to see timeout failures every once in a while during high traffic periods, since those increase the chance to use up the max time _almost_ completely. An overloaded system also contributes towards this edge case.

The solution currently implemented in tip uses a lower bound of 1ms, and raises without retries assuming the timing fits. The solution implemented in this patch uses lower bound = max time, retry at least twice ignoring the time limit, blocking up to 2*max time. I deemed it unlikely enough to happen, so that this contract violation is preferable over getting raises immediately.

That being said, I see potential improvements to get closer to the stated goals. If we track the retry count per rdkafka consumer it becomes possible to "hide" the backoff in the individual poll calls. I.e. we could just sleep and return nil (or empty array) on poll when we're in back-off mode. This would always allows us to do retries, but the next poll after a failure must have the full max time available. To keep current behaviour, it would need to raise if the backoff interval is greater than max time. Downsides are potentially complexity since we need to track retries across consumers and calls to poll. There also might be some "sleep only" poll calls where we don't actually poll, but feign an empty response. Additionally this won't solve the "timeout due to very short max time" issue since we don't set a lower bound. I guess the error can be hidden on the first try if it's a timeout one.

Not retrying is not a decent experience right now. If we raise here, the error will bubble up to `Racecar.run` without the user getting a decent chance to specify an error handler. At least in theory we could use the existing `pause_with_exponential_backoff` feature if we also wrap it around reading from Kafka and ditch the individual retries at poll times, but that's definitely a breaking change.

I hope these considerations are somewhat helpful. I'm open for other ideas here, and probably will also get enough time to implement them (within reason).

breunigs:
Is there a chance to move forward on this? Our use case doesn't have any particular requirements, apart from not crashing, so I am happy with any choice here. I took a closer look when this error occurs, and the only one I can see in our systems is `Rdkafka::RdkafkaError: Local: Broker transport failure (transport)`. The interesting bit is that I _don't_ see any log messages for to `(try #{try}): Error for topic subscription #{current_subscription}: #{e}`, so potentially rdkafka waits until the timeout exceeds and raises immediately. I'm not 100% certain on this, since I don't know how to reproduce the issue, or at least would need to spend considerable amount of time to setup a local network with the right brokenness.

dasch:
Black box decoding these issues is tricky, I can say from experience :grimace:

dasch:
How about we just skip the last poll if `remain < max_wait_time`? That is, a precondition for doing the poll is that we have sufficient time to handle it properly.

breunigs:
Thank you for finding the time to look at this.

Unless I am misunderstanding something, that would imply `batch_poll` of doing up to 1 poll in total. Well, ignoring any rounding errors and other "0 time spent" polls edge cases.

Relevant code from master:
```ruby
 # runner.rb
 consumer.batch_poll(config.max_wait_time_ms).group_by(&:partition)
 
 # consumer_set.rb
 def batch_poll(timeout_ms)
      @batch_started_at = Time.now
      @messages = []
      while collect_messages_for_batch? do
        remain = remaining_time_ms(timeout_ms, @batch_started_at)
        break if remain <= 0
        msg = poll(remain)
        break if msg.nil?
        @messages << msg
      end
      @messages
    end

    def collect_messages_for_batch?
      @messages.size < @config.fetch_messages &&
      (Time.now - @batch_started_at) < @config.max_wait_time
    end
```

So, `timeout_ms` is `config.max_wait_time_ms` is `config.max_wait_time` is 1 second by default. That means that `remaining_time_ms` ("time remaining") and `Time.now - @batch_started_at` ("time spent") are essentially the same intent, and apart from rounding errors also make the same comparison. I guess this is just a cleanup that should've been done without any change of the algorithm. I'm going to conflate the two in the following, ignoring any unit differences.

In the first iteration, it'll be "remain = timeout_ms = config.max_wait_time". So "remain < max_wait_time" is false, making a poll. That takes non-zero time, so in the 2nd iteration "remain < max_wait_time" is true this time, exiting the loop already.

Maybe I misunderstood you there and you meant the inverse, i.e. a "only poll if remain < max_wait_time". However, apart from equality this is always going to be true as well. 

Looking at the version presented in this PR, the argument would be largely the same.

dasch:
Argh, this is a bit complicated.

I still think the right approach here is to approach each logical poll with a "budget" of time (`max_wait_time`). _Oftentimes_ you'd get a response before that duration expires, because there's enough data available (`min_bytes`), so `remain` wouldn't be close to zero.

Perhaps a pragmatic approach would be to set a hardcoded `min_remain` value – if `remain` goes under that, don't try another fetch. If it's still above that, pass `remain` as the `max_wait_time` for the subsequent fetch. Would that work?

breunigs:
Sure thing, let's assume for any further discussion we want to treat `max_wait_time` as a hard limit that should not be exceeded under any circumstances and adjust the secondary goals accordingly. I don't mean to criticize this requirement, I think it's a very valid one to have. For our use case max time vs more batching is largely not an issue, so I'm fine with either trade off. For us it's mostly the "raise without retry".

I don't think a `min_remain` will help in the "raise without retry" scenario, since potentially the first fetch will take `max_wait_time` and _then_ raise, completely sidestepping the logic. I expect it to make the scenario much less likely to occur though, so we could try with that change in the field to see if this is "good enough".

To always ensure a retry with the `min_remain` approach requires something like `max_time_was_pass_to_rdkafka = max_wait_time - min_remain - wait_time`, i.e. we ensure that the rdkafka poll exits early enough so that we have enough time for backoff-sleeping and a retry with `min_remain` time. Technically a contract change, because `max_wait_time` is documented as "How long to allow the Kafka brokers to wait before returning messages" and this would leave the _brokers_ with `wait_time` ms less. Then again, I doubt conveying this technical difference to users is feasible or helpful.

Either variant has no effect on fully saturated pipelines (max_message constraint is hit before timeout), or very low volume ones where we don't batch to begin with (message interval is greater than `max_wait_time`). We're reducing batching a bit on pipelines that trickle messages every few ms, because we're not spending our full `max_wait_time`. Again, just pointing this out, I'm fine with this trade off.

Does the "extended variant" sound good to you? If so, then I can amend the PR and think through all the edge cases I might've forgotten.

dasch:
Hmm, changing the semantics of `max_wait_time` in that way will probably be confusing. If we're going down that route we should rather introduce a wholly new concept of "wait time".

Trying to circle back to the beginning: what is the core problem we're trying to solve? Could we perhaps just accept that it's not always possible to retry when there's not enough time left and perhaps just make that work better?

breunigs:
The issue we're observing is `Rdkafka::RdkafkaError: Local: Broker transport failure (transport)`, without any retries. Since the consumer/producer is not involved at this stage, the earliest a user can catch this would be around the `Racecar.run` invocation.

An approach I see that might work without introducing a new option, nor changing the semantics of `max_wait_time` would be to withhold the error if there's no time to do a retry, and essentially just retry on the next `batch_poll` call. Implementation wise we'd need to track if the last batch poll ended in such a situation, so that we will not indefinitely retry. This only is an issue if a single rdkafka-poll takes `max_wait_time` and then errors out. From the top of my head, the downside is mostly when testing this in a repl and this edge case occurs, because you'd only see the error on the 2nd try. Since I wasn't able to reproduce it, it won't happen often, though, and if it does printing a warning will mitigate this.

breunigs:
I think it makes sense to let it crash eventually, especially if it encounters the same error. If it's an unrecoverable one, a restart might fix it. At the very least it works as a sort of escalation if users don't setup additional alerts and make this visible in the orchestrator. I'll ensure a metric is exposed for this either way, or increment one of the existing ones (not up to date on the current state there). However, I'd still suggest to let it crash if the same errors occurs, say, for 5*`max_wait_time`.

dasch:
:+1: give it a shot

breunigs:
@dasch want to take another look? The approach now honors the max time limit, of course still split with wait/sleep and giving some time to rdkafka for polling, as before. As for the "crash eventually", it just tries up to MAX_POLL_TRIES, capping the exponential back off if needed. So for an unrecoverable error and rdkafka always using up the given time, you'd get something like:
```
batch_poll(1s)
rdkafka.poll(1s)
return []

batch_poll(1s)
sleep 100ms
rdkafka.poll(900ms)
return []

batch_poll(1s)
sleep 200ms
rdkafka.poll(800ms)
return []

…

batch_poll(1s)
sleep 999ms
rdkafka.poll(1ms)
return []

…

batch_poll(1s)
sleep 999ms
rdkafka.poll(1ms)
raise
```

dasch (approved):
Awesome! Thanks for the contribution!

dasch:
Can you add a line to the changelog as well?

dasch:
Thanks!

PR #178, opened 2020-08-12
Adding support to OAUTHBEARER authentication (sasl_oauth_token_provider)

rafaelbiriba:
Based on https://github.com/zendesk/ruby-kafka/tree/master#oauthbearer

I am adding the sasl_oauth_token_provider configuration, so we can use the oauthbearer from ruby-kafka.

--- 

```
Racecar.configure do |config|
  config.brokers = "localhost:9094"
  config.client_id =  "client"
  config.sasl_oauth_token_provider =  TokenProvider.new
  config.ssl_ca_certs_from_system = true
end
```

```
I, [2020-08-12T10:43:55.007741 #51439]  INFO -- : New topics added to target list: my-topic-name
I, [2020-08-12T10:43:55.007852 #51439]  INFO -- : Fetching cluster metadata from kafka://localhost:9094
D, [2020-08-12T10:43:56.196415 #51439] DEBUG -- : Obtaining new OAuth token
I, [2020-08-12T10:43:58.037736 #51439]  INFO -- : Discovered cluster metadata; nodes: localhost:9094 (node_id=0)
I, [2020-08-12T10:43:58.038700 #51439]  INFO -- : [[client] {}:] Joining group `group_name`
I, [2020-08-12T10:43:58.039412 #51439]  INFO -- : [[client] {}:] Will fetch at most 1048576 bytes at a time per partition from my-topic-name
I, [2020-08-12T10:43:58.039662 #51439]  INFO -- : [[client] {}:] Fetching cluster metadata from kafka://localhost:9094
I, [2020-08-12T10:43:58.039454 #51439]  INFO -- : [[client] {}:] Fetching cluster metadata from kafka://localhost:9094
I, [2020-08-12T10:43:59.677859 #51439]  INFO -- : [[client] {}:] Discovered cluster metadata; nodes: localhost:9094 (node_id=0)
I, [2020-08-12T10:43:59.678802 #51439]  INFO -- : [[client] {}:] There are no partitions to fetch from, sleeping for 1.0s
```

### DIFF
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -4,9 +4,6 @@ module Racecar
   class Config < KingKonf::Config
     env_prefix :racecar
 
-    desc "The OAUTHBEARER token provider class"
-    attr_accessor :sasl_oauth_token_provider
-
     desc "A list of Kafka brokers in the cluster that you're consuming from"
     list :brokers, default: ["localhost:9092"]
 
@@ -143,6 +140,9 @@ module Racecar
     attr_reader :error_handler
 
     attr_accessor :subscriptions, :logger
+
+    # The OAUTHBEARER token provider class.
+    attr_accessor :sasl_oauth_token_provider
 
     def initialize(env: ENV)
       super(env: env)
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -103,6 +103,9 @@ module Racecar
     desc "The SCRAM mechanism to use, either `sha256` or `sha512`"
     string :sasl_scram_mechanism, allowed_values: ["sha256", "sha512"]
 
+    desc "The OAUTHBEARER token provider class"
+    string :sasl_oauth_token_provider
+
     desc "Whether to use SASL over SSL."
     boolean :sasl_over_ssl, default: true
 
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -4,6 +4,9 @@ module Racecar
   class Config < KingKonf::Config
     env_prefix :racecar
 
+    desc "The OAUTHBEARER token provider class"
+    attr_accessor :sasl_oauth_token_provider
+
     desc "A list of Kafka brokers in the cluster that you're consuming from"
     list :brokers, default: ["localhost:9092"]
 
@@ -102,9 +105,6 @@ module Racecar
 
     desc "The SCRAM mechanism to use, either `sha256` or `sha512`"
     string :sasl_scram_mechanism, allowed_values: ["sha256", "sha512"]
-
-    desc "The OAUTHBEARER token provider class"
-    string :sasl_oauth_token_provider
 
     desc "Whether to use SASL over SSL."
     boolean :sasl_over_ssl, default: true
--- a/README.md
+++ b/README.md
@@ -319,11 +319,12 @@ If using SCRAM:
 
 If using OAUTHBEARER:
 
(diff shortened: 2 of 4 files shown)


RaVbaker:
As I can see it only works for version 1. Since v2 of Racecar is based on different kafka client library ([rdkafka-ruby](https://github.com/appsignal/rdkafka-ruby)) which doesn't support OAUTHBEARER authentication method yet.

rafaelbiriba:
@RaVbaker this is why I am merging against V1-stable.

lib/racecar/config.rb
```diff
@@ -103,6 +103,9 @@ class Config < KingKonf::Config
     desc "The SCRAM mechanism to use, either `sha256` or `sha512`"
     string :sasl_scram_mechanism, allowed_values: ["sha256", "sha512"]
 
+    desc "The OAUTHBEARER token provider class"
+    string :sasl_oauth_token_provider
```
eldawy-blacklane, line 107: so type here is `string` but you're sending a `class` did you test if this is working? could you also provide an example?

rafaelbiriba: Yes, it does work!

```
Racecar.configure do |config|
  config.brokers = "localhost:9292"
  config.client_id =  "client"
  config.sasl_oauth_token_provider =  KafkaTokenProvider.new
  config.ssl_ca_certs_from_system = true
end
```

rafaelbiriba:
@dasch do you mind to take a look?
Would be nice to have it merged soon!

lib/racecar/config.rb
D-side-BL: The config is based on [KingKonf](https://github.com/dasch/king_konf) which is designed around the idea of pure data configuration values overridable via env vars or YAML. A token class isn't exactly data, so it might be best to confine the possibility to define it to just runtime (not a huge fan of poking holes through abstractions like specifying a class name, but maybe that's just me). Say, by making it just an `attr_accessor` on the config class.
It would also lose `desc`ription that way... but I don't see a better yet still simple way around this yet.

dasch: Does this work when setting the config in an ENV var or the config file?

rafaelbiriba: env var only, since it required a class instance on it. Should I add this to the readme?

dasch: I agree with @D-side-BL on this – it would be preferable to use an accessor on the Config class instead, since this setting doesn't follow the normal rules for KingKonf config.

dasch: Either that, or using e.g. `get_const` to find the class based on a string config setting.

rafaelbiriba: ok I will change to `accessor` and I will add to the readme that this option only works for ruby config file.

rafaelbiriba: @dasch  can you check again please?

rafaelbiriba:
Thanks for merging it! Can you generate a new version of it?

rafaelbiriba:
@dasch Thanks for merging it! Can you generate a new version of it?

rafaelbiriba:
Hey @dasch , sorry to bother!
Can you generate a new version of it?
We want to deploy it to production with a proper tag/version on it.

rafaelbiriba:
Hey @dasch , sorry to bother!
Can you generate a new version of it?
We want to deploy it to production with a proper tag/version on it.

dasch:
:+1: released in v1.2.0.

PR #179, opened 2020-08-18
Fix max_wait_time usage

edgurgel:
Hi team! 

While checking the Statsd metrics that were being sent we noticed that the `consumer.loop.duration` was being called very frequently and apparently not respecting the `max_wait_time` configuration.

After applying the change from this PR we got the expected result! 

I think I've found the places where the configuration was being sent as seconds but they were expected to be in milliseconds:

https://github.com/zendesk/racecar/blob/c9b50bac4830e6b4f8a25a73cedaff5653c0e9d9/lib/racecar/consumer_set.rb#L16

I added this method to `Racecar::Config` to convert it, but happy to do it differently if there's a better way.

I'm also not sure how to write a good test for this change... any suggestions?

Thanks in advance!

### DIFF
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -156,6 +156,10 @@ module Racecar
 
     attr_accessor :subscriptions, :logger
 
+    def max_wait_time_ms
+      max_wait_time * 1000
+    end
+
     def initialize(env: ENV)
       super(env: env)
       @error_handler = proc {}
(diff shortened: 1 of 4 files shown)


dasch:
Thanks!

dasch:
I've released v2.0.0.beta6. Can you deploy that to production and report back that there are no regressions? Then I think I'm ready to releave v2 final.

edgurgel:
Sure! We will upgrade the version and report back 👍

Issue #186, opened 2020-08-31
racecar will silently crash if an exception is raised from Cli#new


ColinDKelley:
## Problem Statement

`exe/racecar` silently discards exceptions because [it assumes `Cli#run` has already rescued and logged them](https://github.com/zendesk/racecar/blob/master/exe/racecar#L16-L17).
```ruby
rescue
  # Exceptions are printed to STDERR and sent to the error handler
  # in `Racecar::Cli#run`, so we don't need to do anything here.
  exit(1)
```

But if an exception is raised in `Cli#new`, this assumption won't be true. The exception will be silently discarded.

## Steps to Reproduce
Cause an exception to be raised from `Cli#new`. For example, cause a typo in a `--require`:
```sh
$ bundle exec racecar --require typo ConsumerClass || echo failed
```
When you run the above, you will see `failed` output but not the exception that caused it.
```sh
failed
```

## Expected Behavior
When the above is run, the exception should be displayed on stderr (and the error_handler run, if one has been registered):
```sh
=> Crashed: NameError: undefined local variable or method `typo' for Cli#...
<backtrace>

failed
```

## Suggested Fix
The easiest and most general fix for this would be to move the rescue block from [here](https://github.com/zendesk/racecar/blob/master/exe/racecar#L16-L17) out to `exe/racecar`. This is best practice in case some other code path is later introduce that might also `raise`. And it would would have an added benefit of removing any risk of redundantly logging, then raising, then logging the same exception again.

I'd be happy to make a PR with the suggested change if you'd like.

dasch:
Please open a PR and I'll review it!

ColinDKelley:
@dasch Will do!

ColinDKelley:
PR: https://github.com/zendesk/racecar/pull/189

PR #189, opened 2020-09-18
issue #186: move rescue to exe/racecar

ColinDKelley:
- Added `spec/exe/racecar_spec`.
- Moved `rescue` out to `exe/racecar`.
- Added `"--- Caused by: ---\n"` when exception has cause(s).
- Note: moved `exe/racecar` code inside `Racecar.start` to make it possible to stub `exit` without affecting surrounding rspec when it calls `exit`.
- Added `require concurrent/utility/monotonic_time` because that test was failing due to the lack of that require on ActiveSupport 6.0.3.3.
- Moved `config` to `private`.
- Added `pry` as a development dependency.

### DIFF
--- a/lib/racecar/instrumenter.rb
+++ b/lib/racecar/instrumenter.rb
@@ -11,6 +11,7 @@ module Racecar
       @default_payload = default_payload
 
       @backend = if defined?(ActiveSupport::Notifications)
+        require 'concurrent/utility/monotonic_time'
         ActiveSupport::Notifications
       else
         NullInstrumenter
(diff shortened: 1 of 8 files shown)


lib/racecar/instrumenter.rb
```diff
@@ -11,6 +11,7 @@ def initialize(default_payload = {})
       @default_payload = default_payload
 
       @backend = if defined?(ActiveSupport::Notifications)
+        require 'concurrent/utility/monotonic_time'
```
dasch, line 14:
```suggestion
        # ActiveSupport needs `concurrent-ruby` but doesn't `require` it.
        require 'concurrent/utility/monotonic_time'
```

dasch:
Can you update the CHANGELOG?

dasch (approved):

ColinDKelley:
@dasch Thanks for the review. I added the suggested comment, updated the CHANGELOG, and rebased off the latest `master`.

dasch:
Thank you!

PR #193, opened 2020-09-30
Require racecar/version

ellimist:
Require `racecar/version` so we can easily check the currently running version of Racecar in the console.

Currently it doesn't work:

image


dasch:
Thanks! :+1:

Issue #194, opened 2020-10-07
Memory leaks after consumers are not properly "closed" when the broker disconnects from them


Hammadk:
First, thanks for maintaining this lovely gem. After upgrading from Racecar `1.1.0` to  `2.0.0`, we noticed that our Kafka consumers were consistently leaking memory and eventually getting killed because out-of-memory errors. We have identified the root cause.

#### Problem

All of these memory leaks start after a broker disconnects from our consumer:

* [Racecar::ConsumerSet#poll](https://github.com/zendesk/racecar/blob/d095104096ee62894a3d2bd47b11821f1b43abee/lib/racecar/consumer_set.rb#L18-L24) calls [RdKafka::Consumer#poll](https://github.com/appsignal/rdkafka-ruby/blob/532afaa12aab6341bf4410324dee90055d9db1b6/lib/rdkafka/consumer.rb#L421-L442). After a broker disconnects, `RdKafka::Consumer#poll` raises `Local: Broker transport failure (transport)`, which means that this RdKafka error has code `-195` (Rdkafka::RdkafkaError.new(-195))
* Racecar attempts to retry this error but it treats [transport failure errors](https://github.com/zendesk/racecar/blob/d095104096ee62894a3d2bd47b11821f1b43abee/lib/racecar/consumer_set.rb#L33-L35) differently than other errors
* [Racecar::ConsumerSet#reset_current_consumer](https://github.com/zendesk/racecar/blob/d095104096ee62894a3d2bd47b11821f1b43abee/lib/racecar/consumer_set.rb#L144-L146) is called and it removes a reference of the effected consumer from the array of consumers with `@consumers[@consumer_id_iterator.peek] = nil`. 

This only removes the reference but we need to [close](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/consumer.rb#L22-L29) the consumer first by calling `RdKafka::Consumer#close`. If they are not closed and just discarded, a [new thread is created under the hood](https://user-images.githubusercontent.com/129764/95362626-580b7b80-089c-11eb-9671-35c652a7ca4e.png) when ConsumerSet creates a new RdKafka consumer.

#### Proposed Solution

Update [Racecar::ConsumerSet#reset_current_consumer](https://github.com/zendesk/racecar/blob/d095104096ee62894a3d2bd47b11821f1b43abee/lib/racecar/consumer_set.rb#L144-L146) so it closes the RdKafka consumer. This is the solution that we implemented in our local patch.

[...]

PR #196, opened 2020-10-14
[Bugfix] Close RdKafka consumer in ConsumerSet#reset_current_consumer to prevent memory leak

Hammadk:
As described in https://github.com/zendesk/racecar/issues/194, this PR updates [Racecar::ConsumerSet#reset_current_consumer](https://github.com/zendesk/racecar/blob/d095104096ee62894a3d2bd47b11821f1b43abee/lib/racecar/consumer_set.rb#L144-L146)  so we [close](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/consumer.rb#L22-L29) the underlying RdKafka consumer before removing iits reference in the `@consumers` array. Doing this prevents a memory leak when the broker disconnects from a consumers.

cc: @dasch

### DIFF
--- a/lib/racecar/consumer_set.rb
+++ b/lib/racecar/consumer_set.rb
@@ -142,7 +142,12 @@ module Racecar
     end
 
     def reset_current_consumer
-      @consumers[@consumer_id_iterator.peek] = nil
+      current_consumer_id = @consumer_id_iterator.peek
+      @logger.info "Resetting consumer with id: #{current_consumer_id}"
+
+      consumer = @consumers[current_consumer_id]
+      consumer.close unless consumer.nil?
+      @consumers[current_consumer_id] = nil
     end
 
     def maybe_select_next_consumer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* [Bugfix] Close RdKafka consumer in ConsumerSet#reset_current_consumer to prevent memory leak (#196)
+
 ## racecar v2.1.0
 
 * Bump rdkafka to 0.8.0 (#191)
--- a/spec/consumer_set_spec.rb
+++ b/spec/consumer_set_spec.rb
@@ -275,12 +275,13 @@ RSpec.describe Racecar::ConsumerSet do
       consumer_set.send(:select_next_consumer)
     end
 
-    it "#reset_current_consumer does what it says" do
+    it "#reset_current_consumer removes the reference to the Rdkafka consumer" do
       3.times do
         consumer_set.current
         consumer_set.send(:select_next_consumer)
       end
       consumer_set.send(:select_next_consumer)
+      allow(rdconsumer2).to receive(:close)
 
       expect do
         consumer_set.send(:reset_current_consumer)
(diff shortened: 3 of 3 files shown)


lib/racecar/consumer_set.rb
```diff
@@ -142,7 +142,12 @@ def current_subscription
     end
 
     def reset_current_consumer
-      @consumers[@consumer_id_iterator.peek] = nil
+      current_consumer_id = @consumer_id_iterator.peek
+      @logger.info "Resetting consumer with id: #{current_consumer_id}"
+
+      consumer = @consumers[current_consumer_id]
+      consumer.close unless consumer.nil?
```
dasch, line 149: Can `#close` raise an exception? Perhaps we should rescue and log any Rdkafka exception raised here rather than crash the process.

dasch (changes requested):
Small change requested, unless there's reason to believe no exception can be raised there.

Also, can you update the CHANGELOG?

lib/racecar/consumer_set.rb
Hammadk: I see your point about catching an exception because `#reset_current_consumer` is called after exceptions so that raises the chances of unpredictable behavior. However, [#close](https://github.com/appsignal/rdkafka-ruby/blob/532afaa12aab6341bf4410324dee90055d9db1b6/lib/rdkafka/consumer.rb#L22-L29) does not raise an exception as far as I can tell:
* The docs in [Rdkafka::Consumer#close](https://github.com/appsignal/rdkafka-ruby/blob/532afaa12aab6341bf4410324dee90055d9db1b6/lib/rdkafka/consumer.rb#L22-L29) do not indicate that this method raises anything. The docs for other methods like [poll](https://github.com/appsignal/rdkafka-ruby/blob/532afaa12aab6341bf4410324dee90055d9db1b6/lib/rdkafka/consumer.rb#L414-L420) do show that they raises can error.
* `Rdkafka::Consumer#close` calls [rd_kafka_consumer_close](https://github.com/edenhill/librdkafka/blob/dbafbb748c116b03a18b42a10b845eeb5517d03e/src/rdkafka.h#L3660-L3675) and that always returns an integer even on fatal errors
* `Rdkafka::Consumer#close` also calls calls [rd_kafka_destroy](https://github.com/edenhill/librdkafka/blob/dbafbb748c116b03a18b42a10b845eeb5517d03e/src/rdkafka.h#L2699-L2714) which could `trigger consumer callbacks, such as rebalance_cb` which makes this less predictable.

Given that `#close` does raises `Rdkafka::RdkafkaError` we would have to rescue `StandardError` instead. Does that make sense? 

P.s., I found [these librdkafka docs ](https://github.com/edenhill/librdkafka/blob/acca6ed2f96b4a49d13a801e0dae8d373dd847a1/INTRODUCTION.md#high-level-kafkaconsumer) which describes why not not closing the consumer caused the memory leak.

dasch: Can we rescue `Rdkafka::RdkafkaError` or were you saying that `#close` does _not_ raise that?

Hammadk: `#close` does not raise `Rdkafka::RdkafkaError` so I don't think we need to catch it.

dasch: :+1:

dasch (approved):
Thanks!

PR #187, opened 2020-09-07
Keep track of the number of retries when handling errors

cassiomarques:
This PR introduces a new `retries_count` key in the `info` hash passed to the configured error handler.

This can be useful in cases where we might want to limit the number of individual error occurrences that get reported to a exception notification service or even implement some sort of monitoring for errors.

The implementation takes advantage of the existent `Pause` class, which already keeps track of the number of times the runner had to pause for a given topic/partition.

Issue #188, opened 2020-09-15
Allow processing different partitions concurrently


dasch:
While messages from a single partition must be processed in-order, messages from different partitions need not. Therefore, we can use multiple threads to process messages from different partitions.

dasch:
I'm hoping that we can leave the consumer API as-is, and just have this be a configuration. Since both `process` and `process_batch` work on single-partition data, we should be able to spin up _n_ independent consumer class instances and have _n_ threads deliver messages to the processing methods independently.

apeacock1991:
I'm guessing no, but would this parallelise a consumer that subscribes to two topics, each with a single partition?

I was hoping if you had 2 workers, it would assign one topic to one worker, and one topic to another worker 😅

PR #202, opened 2020-11-10
Expose retries_count on Racecar::Message

Alexander-Blair:
In a recent [PR](https://github.com/zendesk/racecar/pull/187), support was added to provide `retries_count` to the instrumentation payload, in order to define custom behaviours on error handling. However, this does not allow us to drive the consumer's behaviour based on the retries, if required.

Example usage:
- We're consuming from a topic where we don't absolutely need to process every single update successfully, however we want to allow for a few intermittent failures
```
class SemiImportantConsumer < Racecar::Consumer
  MAX_RETRIES_COUNT = 5

  subscribes_to "semi_important_topic"

  def process(message)
    data = JSON.parse(message.value)

    do_some_stuff(data)
  rescue StandardError => e
    raise e if message.retries_count < MAX_RETRIES_COUNT

    Honeybadger.notify(e)
    Logger.error("Oh no we ran out of retries")
    ... etc
  end
end
```

### DIFF
--- a/lib/racecar/message.rb
+++ b/lib/racecar/message.rb
@@ -6,8 +6,11 @@ module Racecar
   class Message
     extend Forwardable
 
-    def initialize(rdkafka_message)
+    attr_reader :retries_count
+
+    def initialize(rdkafka_message, retries_count: nil)
       @rdkafka_message = rdkafka_message
+      @retries_count   = retries_count
     end
 
     def_delegators :@rdkafka_message, :topic, :partition, :offset, :key, :headers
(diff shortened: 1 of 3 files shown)


dasch (approved):
Makes sense :+1:

Alexander-Blair:
Thanks @dasch ! If you are still interested in looking at the concurrent processing (detailed in https://github.com/zendesk/racecar/issues/188), would be happy to help out where I can

dasch:
@Alexander-Blair that would be great! Let me know if you'd like some pointers. There's also the ruby-kafka Slack if you'd prefer that.

dasch:
(or the new, fancy Discussions here on GH!)

PR #204, opened 2020-11-13
changed log level from error to warn in poll_with_retries

daviduebelacker:
If I understand it correctly, this is not really a critical error, but can happen from time to time.

For our ops it would be great if we could reduce the loglevel to warn here.


dasch (approved):
Makes sense.

dasch:
@daviduebelacker can you try rebasing on latest master? Maybe that fixes the build.

deepredsky:
Merged with spec fixes in https://github.com/zendesk/racecar/pull/305

PR #205, opened 2020-11-30
Update king_konf runtime dependency to version 1.0

alejandroperea:
Taking into a look into the last commits of `king_konf` gem, it shouldn't be a problem updating this runtime dependency in racecar, so, this PR updates it to version 1.0


dasch (approved):
Thanks!

PR #209, opened 2020-12-15
Handle ERR_NOT_COORDINATOR

ellimist:
Whenever a "Not Coordinator for group" error gets thrown the consumers
get stuck and they never recover. We should be retrying once that
happens.


ellimist:
I could not validate the `:not_coordinator` symbol for the error code, since I couldn't find where it gets generated from. The error code in librdkafka = 16

PR #210, opened 2020-12-15
Add a new workflow for running integration specs

dasch:


.github/workflows/integration-specs.yml
```diff
@@ -0,0 +1,76 @@
...
+          - 9092:9092
+          - 29092:29092
+        env:
+          KAFKA_BROKER_ID: 1
+          KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+          KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:29092,PLAINTEXT_HOST://localhost:9092
+          KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
+          KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
+          KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+
```
bquorning, line 27: Perhaps you need a health check after starting each service, before running the tests? You can pass arguments to [docker create](https://docs.docker.com/engine/reference/commandline/create/#options) using the [options](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idoptions) key – e.g.
```yaml
        options: --health-cmd "curl something --output /dev/null --silent --fail" --health-interval 10s --health-timeout 5s --health-retries 5
```

.github/workflows/integration-specs.yml
```diff
@@ -0,0 +1,76 @@
...
+    - uses: zendesk/checkout@v2
+    - name: Set up Ruby
+      uses: zendesk/setup-ruby@v1.58.0
+      with:
+        ruby-version: 2.7
+    - name: Build and test with RSpec
+      env:
+        RACECAR_BROKERS: localhost:9092
+      run: |
+        gem install bundler --no-document
```
bquorning, line 38: Actually the `setup-ruby` step will make sure Bundler is installed. And if you add your Gemfile.lock to the repository, it will make sure to install the same version of Bundler that you used locally.

If you add `bundler-cache: true` to the `with` part of setup-ruby, it will run `bundle install` and store the result in a cache (or fetch, if it already exists).

Read more at https://github.com/ruby/setup-ruby#usage

.github/workflows/integration-specs.yml
dasch: Thanks! That did the trick! Took me a while to find a good health check command, though, but `kafka-topics` seems to do the trick and is bundled with the image :+1:

.github/workflows/integration-specs.yml
dasch: Thanks! I'll try to optimize that in a follow-up PR.

PR #212, opened 2020-12-16
Better CI bundle cache

bquorning:
By adding Gemfile.lock to the repository, and `bundler-cache: true` is set in the CI configuration, the `zendesk/setup-ruby` action will take care of installing the gems too. 

Please notice the speed difference in [latest master build](https://github.com/zendesk/racecar/actions/runs/426507907) and [this PR build](https://github.com/zendesk/racecar/actions/runs/426775450).

If checking in Gemfile.lock is a concern – e.g. do we forget to test for newer versions of our dependencies – we might want to let [Dependabot](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically) keep Gemfile.lock updated?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,7 @@ jobs:
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true
     - name: Build and test with RSpec
-      run: |
-        bundle install --jobs 4 --retry 3
-        bundle exec rspec --format documentation --require spec_helper --color --exclude-pattern='spec/integration/*_spec.rb'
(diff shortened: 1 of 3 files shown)


dasch (approved):
Thanks! :+1:

How does one go about enabling Dependabot?

bquorning:
> How does one go about enabling Dependabot?

Their documentation could use some improvement, but I think https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-and-disabling-version-updates#enabling-dependabot-version-updates is the best guide. I think it’s just about adding a `.github/dependabot.yml` file and adding some configuration in there.

dasch:
Ah. Maybe I'll get around to that sometime.

PR #214, opened 2021-01-11
set correct default values for max_wait_time on poll and batch_poll

breunigs:
If the higher level runner Racecar::Runner was used, this makes no
difference, as it already passes the correct @config.max_wait_time_ms
value. If someone uses ConsumerSet directly, without specifying
these values, this will be a breaking change.


dasch (approved):
Thanks!

PR #215, opened 2021-01-13
silence error about no time left when there were no errors

breunigs:
This occurs when there is only little time left and the remain_ms
works itself out to 0ms. The error message given in this scenario
before this patch is unhelpful or misleading, and doesn't need to
be acted upon most of the time. The edge case where it matters is
if remain_ms is always 0, e.g. if the system is too slow for the
configured max_wait_time_ms. Hence the debug message to see this
more easily, though it is expected that this shows up on healthy
and properly configured systems, too.

Related to https://github.com/zendesk/racecar/pull/204 . This PR
should make the error much less likely to appear in the first place.
The test included in this PR will fail if the other PR is merged,
because it checks using the log level of the error message (it's an
easy fix, though).


dasch:
Can you rebase on master in that case?

dasch:
Thanks!

PR #219, opened 2021-02-03
reset producer on unrecoverable message delivery errors (fixes #218)

### LINKED ISSUE #218: producer never recovers after network split
I have no good steps to reproduce, especially since this requires breaking the network (or killing the broker) in the right way and is potentially timing related. I'm especially hazy on how to get into this state, though I'm fairly certain on the effects this causes in racecar. I'll provide the investigated variant, though I have more production logs available on demand.

Roughly:
1. racecar is up and operating fine
2. introduce a network split to lose connection to all brokers. Everything will be considered down from librdkafka point of view.
3. eventually restore connection again
4. racecar will clear stuck rdkafka consumers through [`ConsumerSet#reset_current_consumer`](https://github.com/zendesk/racecar/blob/61afeca4bd3ed78eb65d194b996257c424dc7642/lib/racecar/consumer_set.rb#L183) in certain scenarios. This was enough to eventually receive messages again.
5. publish a message (in reaction to the just received ones). Adding it to librdkafka's queue will work. However, it will fail in [`Consumer#deliver!`](https://github.com/zendesk/racecar/blob/61afeca4bd3ed78eb65d194b996257c424dc7642/lib/racecar/consumer.rb#L50) since there is no broker connection, so rdkafka eventually throws a [`WaitTimeoutError`](https://github.com/appsignal/rdkafka-ruby/blob/532afaa12aab6341bf4410324dee90055d9db1b6/lib/rdkafka/abstract_handle.rb#L52). Note that the method exits before calling `@delivery_handles.clear`. This is fine, a handle can be waited upon multiple times.
6. The error bubbles into the pause handling, pausing this topic-partition for some time.
7. eventually we will receive another message; doesn't matter which partition it is from. Eventually it will try to `@delivery_handles.each(&:wait)` again, which waits on the first message described in step 5.
8. after enough retries, and once `message.timeout.ms` is exceeded, librdkafka will throw [`msg_timed_out`](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#error-local-time-out) which means that it has given up. The message will not be sent. 

At the very least, racecar should remove these finally failed messages/handles from `@delivery_handles`. Even if the producer eventually recovers, these failed messages will result in head of line blocking. Of course, as is today, racecar should still fail on those so that the single message or the whole batch block gets retried. 

That being said, this might not be enough. From the statistics librdkafka makes available, I know that all producer broker connections were in `state: "INIT"`. I still have to check if this is the default and librdkafka only connects on first use, or if it indicates librdkafka was additionally stuck for another reason.

breunigs:


breunigs:
This passed internal review. I am not sure what's up with the builds, but https://github.com/zendesk/racecar/pull/219/commits/51d04ce7b9bf7a6e1134b9ee8aeeb5e684b00149 should probably be cherry-picked onto master

lib/racecar/consumer.rb
```diff
@@ -34,13 +34,22 @@ def subscribes_to(*topics, start_from_beginning: true, max_bytes_per_partition:
...
 
-    # Delivers messages that got produced.
+    # Blocks until all messages produced so far have been successfully published. There
+    # are two scenarios where the function raises:
+    # Rdkafka::AbstractHandle::WaitTimeoutError: reached the default timeout of 60s. You
+    #     can call deliver! again to wait longer.
+    # Rdkafka::RdkafkaError: delivery failed for the reason in the exception. Some of
+    #     them include permanent failures, e.g. "msg_timed_out" which librdkafka sets
+    #     after failing to deliver the message within "message.timeout.ms". The caller
+    #     must handle these edge cases or run into head of line blocking.
```
dasch, line 52: Would you be up for mapping the Rdkafka exceptions to Racecar native exceptions with good error messages and UX?

lib/racecar/runner.rb
```diff
@@ -214,6 +219,26 @@ def process_batch(messages)
       end
     end
 
+    # librdkafka will continue to try to deliver already queued messages, even if ruby-rdkafka
+    # raised before that. This method detects any unrecoverable errors and resets the producer
+    # as a last ditch effort.
+    # The function returns true if there were unrecoverable errors, or false otherwise.
+    def reset_producer_on_unrecoverable_delivery_errors(error)
+      return false unless error.is_a?(Rdkafka::RdkafkaError)
+      return false unless error.code == :msg_timed_out # -192
```
dasch, line 228: So this code _only_ handles timeouts, or does `msg_timed_out` cover any reason for error?

dasch (approved):
Looks good! I would _love_ for these errors to be wrapped at the Racecar level, though, with clear clear messages and instructions on what to do. If you're not up for that I'm fine with merging this as-is, though.

lib/racecar/runner.rb
breunigs: The timeout is a variant of permanent failure, [in upstream's words](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#error-local-time-out):
> The message could not be successfully transmitted before message.timeout.ms expired, typically due to no leader being available or no broker connection. The message may have been retried due to other errors but those error messages are abstracted by the ERR__MSG_TIMED_OUT error code.

Meaning that librdkafka does not report any transient errors on the delivery callback. As per [this comment](https://github.com/edenhill/librdkafka/issues/1066#issuecomment-280776570), we have three+one scenarios:
1. success
2. librdkafka gave up due to whatever reason.
3. broker rejects (e.g. message too large), [see upstream docs](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#error-permanent-errors).
4. `Rdkafka::AbstractHandle::WaitTimeoutError` is introduced by rdkafka-ruby, and returns after 60s iirc, even if the failure is not final (i.e. scenario 2)

1 is obviously a no op, 3 is unrecoverable unless the client/caller/implementer changes their message, 2 is this PR. Not all errors that cause 2 can be solved by reconnecting, but a stuck broker connection might be fixed that way. 4 is potentially something racecar can just hide (as in `log.Info "still trying to deliver message..."`) until 2 occurs.

It should be easy enough to wrap these with an extended error message, pointing to the relevant upstream docs and kafka or racecar settings.

For `Rdkafka::AbstractHandle::WaitTimeoutError` the most sensible options seems to be to not have the error message in the first place on high level consumers by setting a wait time > `message.timeout.ms`. The other option would be to try to log something helpful like the message key, but then again we suspect it's not specifically that message but something "on infrastructure level". Plus printing the key if it's a megabyte of JSON also has its own set of problems. So the only reliable takeaway would be that the Ruby process is not stuck… which one can have in a better way from [rbspy](https://rbspy.github.io/).

dasch: Awesome! It would also be _amazing_ if you could put this knowledge into the code as a comment, since it's not at all obvious how that works.

breunigs:
force pushed to retain fast-forwardable, sensible commits. I included hints for the errors that I know or could find reasonable extended help for that is more than just what's included in the original error message.

Technically this is a breaking change for anyone relying on rescuing a `Rdkafka::RdkafkaError`. This can be reduced by making `Racecar::MessageDeliveryError` inherit from `Rdkafka::RdkafkaError`, but that will bring lots of unused and potentially misleading fields – only `rdkafka_response` (integer error code) and thus `code` (human readable Ruby symbol error code) are set in the case at hand. Plus of course, the error message necessarily is different anyway. For the high level consumers, this means they will need to adjust the `error_handler`, which seems an okay trade off instead of awkwardly inheriting the rdkafka error. I have no strong opinion on that one, though.

breunigs:
(and, it would be nice to pass `logger` and `config` from the Runner to the Consumer through e.g. `Consumer.configure(…)`, but I kept the interface stable, since I can sneakily access the logger from rdkafka, and the config is only needed to _not_ generate a "still trying to deliver" message. Both seemed more acceptable than changing the interface, or introducing some backwards compatible but ultimately confusing extra options. I feel a bigger refactor is needed to solve this decently, but I don't think we should API break here for so little gain)

lib/racecar/message_delivery_error.rb
```diff
@@ -0,0 +1,100 @@
...
+          Either increase the broker's "log.segment.bytes", or decrease any of the client's related settings "batch.num.messages", "batch.size" or "message.max.bytes". None of these are configurable through Racecar yet, as the defaults should be sufficient and sane.
+
+          Upstream documentation:
+          broker: https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#brokerconfigs_log.segment.bytes
+          client: https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+        EOM
+
+      when :topic_authorization_failed # 29
+        <<~EOM
+          Failed to deliver message because lacking the authorization to write into the topic.
```
dasch, line 83:
```suggestion
          Failed to deliver message because of insufficient authorization to write into the topic (#{@report&.topic}).
```

Is `topic` available on `@report`?

lib/racecar/consumer.rb
```diff
@@ -34,20 +36,46 @@ def subscribes_to(*topics, start_from_beginning: true, max_bytes_per_partition:
...
         instrumentation_payload = { delivered_message_count: @delivery_handles.size }
 
         @instrumenter.instrument('deliver_messages', instrumentation_payload) do
-          @delivery_handles.each(&:wait)
+          @delivery_handles.each do |handle|
+            # rdkafka-ruby checks every wait_timeout seconds if the message was
+            # successfully delivered, up to max_wait_timeout seconds before raising
+            # Rdkafka::AbstractHandle::WaitTimeoutError. librdkafka will (re)try to
+            # deliver all messages in the background, until "config.message_timeout"
+            # (message.timeout.ms) is exceeded. Phrased differently, rdkakfa-ruby's
```
dasch, line 63:
```suggestion
            # (message.timeout.ms) is exceeded. Phrased differently, rdkafka-ruby's
```

dasch (changes requested):
❤️ 

I think it's fine to introduce the new exception class – any code that relied on the Rdkafka exception would be brittle, and I'd rather get this improvement in now. The changelog entry is clear :+1:

I've added a couple of small suggestions but otherwise this looks great! Thanks!

dasch:
> (and, it would be nice to pass `logger` and `config` from the Runner to the Consumer through e.g. `Consumer.configure(…)`, but I kept the interface stable, since I can sneakily access the logger from rdkafka, and the config is only needed to _not_ generate a "still trying to deliver" message. Both seemed more acceptable than changing the interface, or introducing some backwards compatible but ultimately confusing extra options. I feel a bigger refactor is needed to solve this decently, but I don't think we should API break here for so little gain)

`Racecar.config` should be available. Since there's only one Consumer running in a process at a time, using this global singleton should be fine.

dasch:
Can you try rebasing on master and perhaps running bundle install if necessary?

breunigs:
> > (and, it would be nice to pass `logger` and `config` from the Runner to the Consumer through e.g. `Consumer.configure(…)`, but I kept the interface stable, since I can sneakily access the logger from rdkafka, and the config is only needed to _not_ generate a "still trying to deliver" message. Both seemed more acceptable than changing the interface, or introducing some backwards compatible but ultimately confusing extra options. I feel a bigger refactor is needed to solve this decently, but I don't think we should API break here for so little gain)
> 
> `Racecar.config` should be available. Since there's only one Consumer running in a process at a time, using this global singleton should be fine.

`config` works exactly the same as `logger` here. See the other comment for that. So we can decide to do `config`, `logger`, or both actually, though we only need one :)

lib/racecar/consumer.rb
```diff
@@ -34,20 +36,46 @@ def subscribes_to(*topics, start_from_beginning: true, max_bytes_per_partition:
...
+            # The raising can be avoided if max_wait_timeout below is greater than
+            # config.message_timeout, but config is not available here (without
+            # changing the interface).
+            handle.wait(max_wait_timeout: 60, wait_timeout: 0.1)
+          rescue Rdkafka::AbstractHandle::WaitTimeoutError => e
+            delivery_report = handle.create_result
+            # ideally we could use the logger passed to the Runner, but it is not
+            # available here. The runner sets it for Rdkafka, though, so we can use
+            # that instead.
+            Rdkafka::Config.logger.debug "Still trying to deliver message to partition #{delivery_report.partition} at offset #{delivery_report.offset}... (will try up to Racecar.config.message_timeout)"
```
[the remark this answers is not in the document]
dasch: > We can of course reduce the likelihood of that happening by changing the signature of `configure` to `configure(producer:, consumer:, instrumenter: NullInstrumenter, logger: Racecar.config.logger)` and adapting Racecar internal calls to it. Then someone needs to call `configure` on the consumer for this to break.

I think that makes sense :+1:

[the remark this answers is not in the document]
dasch: > > Also... the offset being reported is surely not from the produced message, right?
> 
> It's good that you bring that up, it's very likely always some -1 default value. It _will_ be from the produced message, but most likely only when it actually gets delivered. All these fields get initialized to -1 from rdkafka:
> https://github.com/appsignal/rdkafka-ruby/blob/c036e1ee1173b7547ea194647550592bdfac3469/lib/rdkafka/producer.rb#L122-L126
> It then hands a pointer to the handle to librdkafka which will eventually fill in all the details. Potentially the partition can be filled even if the delivery fails, since partitioning is done client side (at least potentially, librdkafka has code for it and the [on wire protocol](http://kafka.apache.org/0101/protocol.html#protocol_messages) seems to allow it).
> 
> So long story short, the offset is very likely always -1, and I assume the partition will also be. I'll remove all the extra output, since it's very unlikely it will be helpful.

The partition is _always_ set by the client, but it may not be present on that object – although I would guess that it would be. Perhaps keep partition and we can change it later if it's empty in practice?

lib/racecar/message_delivery_error.rb
dasch: People overwriting `produce` are already breaking the implicit contract of Racecar – that's not something you should do. If you're up for it, I think it would make sense to look into a Racecar wrapper for the delivery handles that kept that extra data. Doesn't have to be in this PR, though.

dasch:
Can you try rebasing on master _again_? There was a misconfiguration of CI.

dasch:
Thanks! 💪

PR #222, opened 2021-02-23
Concurrency testing using a forking approach

Alexander-Blair:
The issue: https://github.com/zendesk/racecar/issues/188

Some discussion on a previous implementation approach on this PR: https://github.com/zendesk/racecar/pull/221 - this PR is an initial draft to continue the conversation.

Adds a `ConcurrentRunner`, which uses forking to allow concurrent processing:
- Adds the code in a way that does not affect existing APIs (you never know what your users will decide to depend on)
- Includes quite a big overhaul of the existing integration test so that we have a fresh topic for each spec, and we also clean up the topics at the end.

What hasn't been done yet?
- Extensive testing
- Some kind of benchmarking in a semi-realistic setup. Few possible scenarios:
  - Consuming a pile of messages and writing one straight out again
  - Consuming a pile of messages and syncing it into a database

Outstanding question(s):
- Should there be a hard limit on max concurrency?

### DIFF
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -167,7 +167,7 @@ module Racecar
     # The error handler must be set directly on the object.
     attr_reader :error_handler
 
-    attr_accessor :subscriptions, :logger
+    attr_accessor :subscriptions, :logger, :parallel_workers
 
     def statistics_interval_ms
       if Rdkafka::Config.statistics_callback
@@ -220,6 +220,7 @@ module Racecar
         consumer_class.name.gsub(/[a-z][A-Z]/) { |str| "#{str[0]}-#{str[1]}" }.downcase,
       ].compact.join
 
+      self.parallel_workers = consumer_class.parallel_workers
       self.subscriptions = consumer_class.subscriptions
       self.max_wait_time = consumer_class.max_wait_time || self.max_wait_time
       self.pidfile ||= "#{group_id}.pid"
--- a/lib/racecar/consumer.rb
+++ b/lib/racecar/consumer.rb
@@ -9,7 +9,7 @@ module Racecar
     class << self
       attr_accessor :max_wait_time
       attr_accessor :group_id
-      attr_accessor :producer, :consumer
+      attr_accessor :producer, :consumer, :parallel_workers
 
       def subscriptions
         @subscriptions ||= []
@@ -27,7 +27,15 @@ module Racecar
       # @param additional_config [Hash] Configuration properties for consumer.
       #   See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
       # @return [nil]
-      def subscribes_to(*topics, start_from_beginning: true, max_bytes_per_partition: 1048576, additional_config: {})
+      def subscribes_to(
+        *topics,
+        start_from_beginning: true,
+        max_bytes_per_partition: 1048576,
+        additional_config: {},
+        parallel_workers: nil
+      )
+        self.parallel_workers = parallel_workers
+
         topics.each do |topic|
           subscriptions << Subscription.new(topic, start_from_beginning, max_bytes_per_partition, additional_config)
         end
--- a/lib/racecar/consumer.rb
+++ b/lib/racecar/consumer.rb
@@ -31,11 +31,8 @@ module Racecar
         *topics,
         start_from_beginning: true,
         max_bytes_per_partition: 1048576,
-        additional_config: {},
-        parallel_workers: nil
+        additional_config: {}
       )
-        self.parallel_workers = parallel_workers
-
         topics.each do |topic|
           subscriptions << Subscription.new(topic, start_from_beginning, max_bytes_per_partition, additional_config)
         end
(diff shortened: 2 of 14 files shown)


dasch:
Try rebasing on master to get the tests working again. Also, would you like to extract the improvements to the test suite to a separate PR and get that merged first?

Alexander-Blair:
Very sensible idea! 👍  https://github.com/zendesk/racecar/pull/225

lib/racecar/concurrent_runner.rb
```diff
@@ -0,0 +1,92 @@
...
+    def wait_for_exit(readers)
+      # These readers are listening to the child processes. They will become ready to
+      # read in the following circumstances:
+      # - An exception is raised in the child process (in this case we should initiate
+      #   a shutdown);
+      # - A graceful shutdown was already initiated, and the pipe writer has been closed (in
+      #   which case there is nothing more to do.
+      readable_io = IO.select(readers)
+
+      first_read = (readable_io.first & readers).first.read
```
dasch, line 84: Can you add a comment to this line, it's not obvious what's being done.

lib/racecar/config.rb
```diff
@@ -153,6 +153,9 @@ class Config < KingKonf::Config
     desc "Whether to boot Rails when starting the consumer"
     boolean :without_rails, default: false
 
+    desc "Maximum number of threads to run the application with. Each will spawn its own consumer"
+    integer :max_concurrency, default: 1
```
dasch, line 157: This should be changed to processes. Also, I think we should call this _parallel_ rather than _concurrent_ now that we're not using threads. Alternatively, we could perhaps look at what the configs are called for Unicorn and Resque?

dasch: Perhaps "workers" is better. Also – it's not really a _max_, is it? It's the actual number of workers that will be spun up.

dasch (changes requested):
Getting close! Can you also rebase on master to see if that fixes CI?

lib/racecar/config.rb
Alexander-Blair: Yeah true. What needs to be clear though is that the number of workers * running instances should be less than or equal to the number of partitions in the Kafka topics. Especially if people are running locally and are disappointed when throughput doesn't improve on a single partitioned topic 😄

dasch:
> One of the workers somehow receives a shutdown signal - in this case,
  we do the same as above, initiating termination and waiting for all of
  the workers to shut down

Is there a case to be made for restarting that worker instead, or does that expand the state space too much?

lib/racecar/concurrent_runner.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        if exception_found
+          self.exception = Marshal.load(first_read)
+          terminate_child_processes
+        end
+
+        ready_readers.each(&:close)
+
+        # Recursively wait for the remaining readers
+        wait_for_exit(remaining_readers - ready_readers)
```
dasch, line 115: This risks blowing the stack if we go too deep. Probably iterate instead, e.g. `until remaining_readers.empty? ... end` and update a variable with the remaining readers.

Alexander-Blair:
> Is there a case to be made for restarting that worker instead, or does that expand the state space too much?

I'd maybe like to have this running as an experimental feature, and see if 'quiet' deaths of workers is a common occurrence. From what I can see, exceptions (aside from on startup) are rescued within the `Racecar::Runner` itself, and baked into the retry capabilities. Perhaps if it turns out the consumers can die on their own (and it turns out to be somewhat common), we can explore the restart capabilities.

lib/racecar/concurrent_runner.rb
Alexander-Blair: Indeed, this wasn't needed in the end 👍

dasch (approved):
Looks good, ready for beta testing at least :+1:

Can you fully document in the README and perhaps the code as well, and also add an entry to the changelog?

lib/racecar/consumer.rb
```diff
@@ -27,7 +27,15 @@ def subscriptions
       # @param additional_config [Hash] Configuration properties for consumer.
       #   See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
       # @return [nil]
-      def subscribes_to(*topics, start_from_beginning: true, max_bytes_per_partition: 1048576, additional_config: {})
+      def subscribes_to(
+        *topics,
+        start_from_beginning: true,
+        max_bytes_per_partition: 1048576,
+        additional_config: {},
+        parallel_workers: nil
```
dasch, line 35: I don't think it makes sense to make this configuration part of the `subscribes_to` method – consumer classes can subscribe to multiple topics, but the parallelism is tied to the consumer, not the topic. The class level accessor should suffice; `class MyConsumer < ...; self.parallel_workers = 5; end`.

dasch (changes requested):
Darn close! Can you make the change to the code and also add an entry in the changelog?

dasch (approved):
Looks great! Thanks a bunch for putting in all the work!

I can see that the branch needs to be rebased, but after that I'll happily merge :D

PR #225, opened 2021-02-23
Rework integration tests in preparation for additional test cases

Alexander-Blair:
- Extracts tasks like waiting for a message to arrive into an integration helper module
- Covers a couple of failure scenarios
- Uses different topics for each spec, and for each suite run
- Cleans up the topics at the end. This should mean that a developer running the suite locally doesn't end up with hundreds of topics hanging around!
- We also allow the graceful shutdown to run after the spec run


dasch:
Has this been rebased on latest master?

dasch:
If so, there must be some other reason why the CI workflow doesn't run...

dasch:
I think #226 fixed the CI issue, can you rebase again?

dasch (approved):
💪

PR #228, opened 2021-02-28
Add Prometheus client integration

fvioz:
The main goal is to provide an easy way to allow scraping metrics from consumers using Prometheus.

Currently, we could use https://github.com/prometheus/client_ruby#pushgateway to export metrics to Prometheus, but this adds an extra layer of complexity in which the code that sends metrics blocks the current consumer, so it forces to add non-blocking code. It also implies that you will start to send a lot of POST request per record consumed.

To solve this, this solution creates a rack server with the Prometheus client in a thread, so it can be scraped by a Prometheus instance.

It also provides an example of how to use ActiveSupport::Subscriber to add some basic metrics instrumented by Racecar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,17 @@ Racecar supports [Datadog](https://www.datadoghq.com/) monitoring integration. I
 * `datadog_namespace` – The namespace to use for Datadog metrics.
 * `datadog_tags` – Tags that should always be set on Datadog metrics.
 
+#### Prometheus monitoring
+
+Racecar supports [Prometheus](https://www.prometheus.io/) monitoring integration. If you are scraping data from other applications, you just need to set `prometheus_enabled` to `true` and have `prometheus-client` gem on your dependencies, as the rest of the settings come with sane defaults.
+
+* `prometheus_enabled` – Whether prometheus monitoring is enabled (defaults to `false`) or not.
+* `prometheus_endpoint` – The server metrics endpoint that is used to fetch prometheus data.
+* `prometheus_registry` – The prometheus client instance.
+* `prometheus_port` – The port use on the server.
+
+After prometheus is enabled, you could instrument your consumers or listen the racecar events with a subscriber as it is in the following [example](examples/racecar_subscriber.rb).
+
 #### Consumers Without Rails ####
 
 By default, if Rails is detected, it will be automatically started when the consumer is started. There are cases where you might not want or need Rails. You can pass the `--without-rails` option when starting the consumer and Rails won't be started.
--- a/examples/racecar_subscriber.rb
+++ b/examples/racecar_subscriber.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class RacecarSubscriber < ActiveSupport::Subscriber
+  attach_to :racecar
+
+  def process_batch(event)
+    counter_by(:processed_message_total).increment(by: event.payload[:message_count], labels: { topic: event.payload[:topic] })
+  end
+
+  def process_message(event)
+    counter_by(:processed_message_total).increment(by: 1, labels: { topic: event.payload[:topic] })
+  end
+
+  def produce_message(event)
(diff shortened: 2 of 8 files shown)


examples/racecar_subscriber.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class RacecarSubscriber < ActiveSupport::Subscriber
```
dasch, line 3: Can we call this e.g. `prometheus_instrumentation.rb`? It's not an example of Racecar in general.

examples/racecar_subscriber.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  def counter_by(name)
+    counter = use_metric(name) { registry.counter(name, docstring: '', labels: [:topic]) }
+  end
+
+  def use_metric(name)
+    registry.exist?(name) ? registry.get(name) : yield
```
dasch, line 25:
```suggestion
    prometheus_registry.exist?(name) ? prometheus_registry.get(name) : yield
```

README.md
```diff
@@ -320,6 +320,17 @@ Racecar supports [Datadog](https://www.datadoghq.com/) monitoring integration. I
 * `datadog_namespace` – The namespace to use for Datadog metrics.
 * `datadog_tags` – Tags that should always be set on Datadog metrics.
 
+#### Prometheus monitoring
+
+Racecar supports [Prometheus](https://www.prometheus.io/) monitoring integration. If you are scraping data from other applications, you just need to set `prometheus_enabled` to `true` and have `prometheus-client` gem on your dependencies, as the rest of the settings come with sane defaults.
+
+* `prometheus_enabled` – Whether prometheus monitoring is enabled (defaults to `false`) or not.
```
dasch, line 327:
```suggestion
* `prometheus_enabled` – Whether prometheus monitoring is enabled (defaults to `false`).
```

README.md
```diff
@@ -320,6 +320,17 @@ Racecar supports [Datadog](https://www.datadoghq.com/) monitoring integration. I
...
+* `prometheus_endpoint` – The server metrics endpoint that is used to fetch prometheus data.
+* `prometheus_registry` – The prometheus client instance.
```
dasch, line 329:
```suggestion
* `prometheus_registry` – The Prometheus client instance.
```

README.md
```diff
@@ -320,6 +320,17 @@ Racecar supports [Datadog](https://www.datadoghq.com/) monitoring integration. I
 * `datadog_namespace` – The namespace to use for Datadog metrics.
 * `datadog_tags` – Tags that should always be set on Datadog metrics.
 
+#### Prometheus monitoring
+
+Racecar supports [Prometheus](https://www.prometheus.io/) monitoring integration. If you are scraping data from other applications, you just need to set `prometheus_enabled` to `true` and have `prometheus-client` gem on your dependencies, as the rest of the settings come with sane defaults.
+
+* `prometheus_enabled` – Whether prometheus monitoring is enabled (defaults to `false`) or not.
+* `prometheus_endpoint` – The server metrics endpoint that is used to fetch prometheus data.
```
dasch, line 328:
```suggestion
* `prometheus_endpoint` – The server metrics endpoint that is used to fetch Prometheus data.
```

README.md
dasch, line 329:
```suggestion
* `prometheus_registry` – The Prometheus client instance.
```

Can you link to some docs for this and perhaps the other concepts? Like, is this a URL for a service or something?

dasch (changes requested):
Instead of the example, can you add a default ActiveSupport::Notifications subscriber to `lib/racecar/prometheus.rb` like the one found in e.g. `datadog.rb`? I think the out-of-the-box setup is what people will want to use in Racecar.

examples/racecar_subscriber.rb
dasch: Actually, I think a real implementation is better than an example.

PR #242, opened 2021-04-22
Bump king_konf to v1.0 in the v1-stable branch

dmitry-semenov:
This change is already in master and it has no breaking changes, I would like to get it into `v1-stable` to keep using with the latest https://github.com/zendesk/delivery_boy which already depends on `"king_konf", "~> 1.0"` (due to the latest `racecar`, so kinda circular dependency).

Thanks in advance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


dasch:
Released v1.2.1.

PR #248, opened 2021-05-19
Native Heroku Kafka support

alanjsph:
Goal is to provide an easy mechanism to connect to Heroku Kafka using the ENVs provided by the broker.

Currently users have to create an initizlizer to read these ENVs and set `Racecar::Config` appropriately. Heroku Kafka provides the certificates as strings, but Racecar (RDKafka) needs them as a files.

To solve this, this solution creates the file `lib/heroku.rb`, which does all the work of parsing the ENVs and setting them to the Config appropriately. Users can just `require 'racecar/heroku'` on top of the Consumer class or in an initializer to make Racecar work with the ENVs provided by Heroku Kafka.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,17 +439,16 @@ The important part is the `strategy.type` value, which tells Kubernetes how to u
 
 Instead, the `Recreate` update strategy should be used. It completely tears down the existing containers before starting all of the new containers simultaneously, allowing for a single synchronization stage and a much faster, more stable deployment update.
 
-#### Deploying to Heroku (Making use of Heroku Kafka ENVs)
+#### Deploying to Heroku (Making use of Heroku Kafka add-on ENVs)
 
-If you run your applications in Heroku and/or use Heroku Kafka as the broker, you will be provided with 4 ENVs to connect to the broker: `KAFKA_URL`, `KAFKA_TRUSTED_CERT`, `KAFKA_CLIENT_CERT`, `KAFKA_CLIENT_CERT_KEY`. Find [relavent Heroku Documentation here](https://devcenter.heroku.com/articles/kafka-on-heroku#connecting-to-a-kafka-cluster).
+If you run your applications in Heroku and/or use Heroku Kafka as the broker, you are be provided with 4 ENVs by the add-on to connect to the broker: `KAFKA_URL`, `KAFKA_TRUSTED_CERT`, `KAFKA_CLIENT_CERT`, `KAFKA_CLIENT_CERT_KEY`.
 
-Racecar provides an easy and flexible way to configure your consumer to work with these ENVs. Just requiring `racecar/heroku` in your consumer file or in one of your rails initializer files will set the Racecar configuration variables to the parsed ENV values.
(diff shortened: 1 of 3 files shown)


dasch:
I didn't realize you could configure the ENV var prefix. I think we should either state in the README that the user needs to manually set up the config in that case, or design this feature differently. Is non-standard prefixes a common thing?

alanjsph:
@dasch Yes, I have covered about how to manually set the Kafka name if it's not the default `KAFKA`.
Also, I don't think having a different name for your Kafka broker (or any Heroku addons for that matter) is non-standard.
It's a feature provided by Heroku to name your addons. This is especially useful when you have 2 same addons attached to an app. And this is very common.

`heroku addons:create heroku-redis:premium1 -a your-app --as REDIS_PRIMARY`
`heroku addons:create heroku-redis:premium1 -a your-app --as REDIS_SECONDARY`

This will give you 2 ENVs - `REDIS_PRIMARY_URL` and `REDIS_SECONDARY_URL`, and you can use them in the app.
It's the same for a Kafka broker instance as well. If you don't mention a name by using the flag `--as`, Heroku will create and attach the add-on by the name/prefix `KAFKA` (If there are no other add-ons in the same name)

alanjsph:
The reason I went with a default name `KAFKA` is that, usually an app will not have not more than one Kafka Brokers attached to it. And the default name provided by Heroku is `KAFKA`. So it's very unlikely that someone would change the attachment name for a `KAFKA`. But it is very common to have named attachments for PG, Redis etc

dasch:
Since I _don't_ think it's common to connect to multiple _Kafka_ clusters from a single Racecar app, I'd prefer to have this "shortcut" work just with the standard prefix – there's nothing stopping people with a more advanced setup from just doing this configuration themselves. The feature is meant as a shortcut that helps with the common case – even having first class support for Heroku only makes sense because it's pretty widely used, but I don't want to take it as far as supporting non-default configurations.

dasch:
I'd think so – I don't see much reason to have multiple Kafka clusters set up for a single application. Does your org have that?

An alternative would be to implement a `Racecar::Heroku.load_configuration!(addon_prefix = "KAFKA")` method, then in the bottom of the file do

```ruby
if ENV.key?("KAFKA_URL")
  # Assume the default addon prefix is used.
  Racecar::Heroku.load_configuration!
end
```

If users don't use the standard prefix, they can call `Racecar::Heroku.load_configuration!(addon_prefix = "MY_KAFKA")` after requiring the file.

alanjsph:
No, we don't have multiple Kafka brokers for a single app.

But it's very common to alias an add-on to a different name which might make more sense to the App, like the `REDIS_PRIMARY` example I mentioned earlier.
With Kafka add-ons though, like you said, it is not very common to have multiple brokers configured for a single app. But that is not stopping the users from having the one single Kafka attached with a different alias. And I don't think this aliasing can be considered as a "non-standard configuration".

But practically, with other add-ons, it is very common to have aliases, especially because we will have same add-ons attached more than once.
With Kafka - it will not be applicable much, because it is not common to have more than one broker attached to an app. But users can still alias the broker, which is a standard Heroku feature. Do people often rename Kafka brokers - I am not sure

dasch:
I'd err on the side of just assuming it's the standard name and see if people report issues.

dasch:
Awesome, thanks!

PR #264, opened 2021-09-23
Add support for dogstatsd-ruby v5

evanlok:
- dogstatsd-ruby v5 buffers metrics before sending them which means they need to be flushed on process shutdown or metrics data may be lost
- The existing `close` method now calls `flush` synchronously https://github.com/DataDog/dogstatsd-ruby/blob/master/lib/datadog/statsd.rb#L331-L334
- This update adds support for dogstatsd-ruby v5 by calling `close` when the process is stopped

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    racecar (2.3.1)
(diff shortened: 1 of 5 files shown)


dasch (approved):
:+1:

PR #266, opened 2021-10-05
Reset consumer on network errors

ronan-mch:


lib/racecar/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Racecar
-  VERSION = "2.4.0"
+  VERSION = "2.4.1"
```
dasch, line 4: Better to leave the version bump out of the PR.

dasch (approved):

jbdietrich:
@ronan-mch is this still relevant or can it be closed?

ronan-mch:
I think we should probably close it. I have completely forgotten the context here.

PR #269, opened 2021-12-03
Allow Consumer (Processor?) to overwrite global `fetch_messages` config.

henrikbjorn:
Consumers might have different batch needs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


dasch:
Thank you! Can you also update the documentation and add an entry in the changelog?

henrikbjorn:
@dasch Done :)

henrikbjorn:
@dasch Is it possible to get a new release cut ? :)

PR #270, opened 2021-12-14
bump rdkafka dependency to 0.11.0

jeenbroekstra:
rdkafka 0.11 uses librdkafka 1.7.0 which has a number of fixes in place (compared to 1.5.0) related to commit sync handling. 

Putting this up for early review and some feedback on whether this is a version update that we can do without risking backward compatibility issues.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     racecar (2.4.0)
       king_konf (~> 1.0.0)
-      rdkafka (~> 0.10.0)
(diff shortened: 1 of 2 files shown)


leonmaia:
Hi @jeenbroekstra thanks for opening the PR, we have just released a [new version of racecar](https://github.com/zendesk/racecar/blob/master/CHANGELOG.md#v270) that uses librdkafka 1.8.2. I'll be closing this PR 👍 .

PR #272, opened 2022-01-01
bugfix: increment attempts counter in integration_helper

sco11morgan:
## What

Increment attempts counter in integration_helper

## Why

Prevent `wait_for_messages` from looping forever if a message is not published


dasch (approved):
Thank you!

PR #275, opened 2022-01-27
Allow producing to a specific partition

codella:
Allow producing to a specific partition.

I have encountered a use case where I want to produce a message (a tombstone, to be specific) to the same partition a consumed message came from.

Unfortunately, there's no guarantee the consumed message was produced using the same partitioner used by Rdkafka, or the one specified by the producing consumer.

Rdkafka already allows to specify a specific partition to produce a message to, so these changes are only passing the partition provided by the Racecar API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add capability to specify partition number when producing messages
+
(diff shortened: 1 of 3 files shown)


dasch (approved):
Can you also leave an entry in the changelog?

PR #306, opened 2022-10-07
Expose partitioner config option

deepredsky:
librdkafka allows different strategies to determine the partitions.
https://docs.confluent.io/5.1.4/clients/librdkafka/CONFIGURATION_8md.html

This PR exposes `partitioner` as an option that can be directly used instead of
using the `producer` config option

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -149,7 +149,9 @@ module Racecar
         "client.id"              => config.client_id,
         "statistics.interval.ms" => config.statistics_interval_ms,
         "message.timeout.ms"     => config.message_timeout * 1000,
+        "partitioner"            => config.partitioner.to_s,
       }
+
       producer_config["compression.codec"] = config.producer_compression_codec.to_s unless config.producer_compression_codec.nil?
       producer_config.merge!(config.rdkafka_producer)
       producer_config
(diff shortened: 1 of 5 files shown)


spec/config_spec.rb
```diff
@@ -40,6 +40,27 @@
     }.to raise_exception(KingKonf::ConfigError)
   end
 
+  it "requires `partitioner` to be a valid" do
```
jbdietrich, line 43:
```suggestion
  it "requires `partitioner` to be a valid value" do
```

lib/racecar/runner.rb
```diff
@@ -148,9 +148,11 @@ def producer_config
         "bootstrap.servers"      => config.brokers.join(","),
         "client.id"              => config.client_id,
         "statistics.interval.ms" => config.statistics_interval_ms,
-        "message.timeout.ms"     => config.message_timeout * 1000,
+        "message.timeout.ms"     => config.message_timeout * 1000
```
dasch, line 151: Let's just keep the trailing commas, they make diffs much nicer and avoids some merge conflicts.

dasch (approved):
Small comments, otherwise :+1:

Can you also update the README and the CHANGELOG?

lib/racecar/runner.rb
deepredsky:
```suggestion
        "message.timeout.ms"     => config.message_timeout * 1000,
```

PR #310, opened 2023-01-04
update File exist method call since exists is deprecated in Ruby 3.2

lynne-ashminov:
File.exists? has been deprecated since Ruby 2.2 and has been removed in Ruby 3.2

Please see: https://bugs.ruby-lang.org/issues/17391

### DIFF
--- a/lib/racecar/daemon.rb
+++ b/lib/racecar/daemon.rb
@@ -54,7 +54,7 @@ module Racecar
     end
 
     def pid
-      if File.exists?(pidfile)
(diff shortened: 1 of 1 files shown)


deepredsky (approved):

deepredsky:
@lynne-ashminov Thanks for fixing the deprecation! could you rebase from origin/master

lynne-ashminov:
> @lynne-ashminov Thanks for fixing the deprecation! could you rebase from origin/master

@deepredsky No problem!  It should be all set :)

PR #311, opened 2023-01-10
Add a standalone Producer to Racecar

pfemo:


mensfeld:
You can plug and play waterdrop: https://github.com/karafka/waterdrop

it seems to have all you described that is needed.

dasch:
@mensfeld this is planned work for Racecar.

lib/racecar.rb
```diff
@@ -39,6 +40,22 @@ def self.logger=(logger)
     config.logger = logger
   end
 
+  def self.produce_async(value:, topic:, **options)
+    producer.produce_async(value: value, topic: topic, **options)
+  end
+
+  def self.produce_sync
+    producer.produce_sync
```
dasch, line 48:
```suggestion
  def self.produce_sync(value:, topic:, **options)
    producer.produce_sync(value: value, topic: topic, **options)
```

lib/racecar/producer.rb
```diff
@@ -0,0 +1,96 @@
...
+      nil
+    end
+
+    # synchronous message production - will wait until the delivery handle succeeds, fails or times out.
+    def produce_sync(value:, topic:, **options)
+      begin
+        handle = internal_producer.produce(payload: value, topic: topic, **options).wait(max_wait_timeout: 60, wait_timeout: 0.1)
+      rescue Rdkafka::AbstractHandle::WaitTimeoutError => e
+        partition = MessageDeliveryError.partition_from_delivery_handle(handle)
+        @logger.debug "Still trying to deliver message to (partition #{partition})... (will try up to Racecar.config.message_timeout)"
```
dasch, line 34:
```suggestion
        @logger.warn "Still trying to deliver message to (partition #{partition})... (will try up to Racecar.config.message_timeout)"
```

I think this is significant enough, and should happen rarely enough, that it warrants `WARN`.

lib/racecar/producer.rb
```diff
@@ -0,0 +1,96 @@
...
+        @delivery_handles.each do |handle|
+          begin
+            handle.wait
+          rescue Rdkafka::AbstractHandle::WaitTimeoutError => e
+            partition = MessageDeliveryError.partition_from_delivery_handle(handle)
+            @logger.debug "Still trying to deliver message to (partition #{partition})... (will try up to Racecar.config.message_timeout)"
+            retry
+          rescue Rdkafka::RdkafkaError => e
+            raise MessageDeliveryError.new(e, handle)
+          end
```
dasch, line 69: We can avoid this duplication using a helper method, e.g.

```ruby
private
def deliver_with_error_handling(handle)
  handle.wait
rescue Rdkafka::AbstractHandle::WaitTimeoutError => e
  partition = MessageDeliveryError.partition_from_delivery_handle(handle)
  @logger.debug "Still trying to deliver message to (partition #{partition})... (will try up to Racecar.config.message_timeout)"
  retry
rescue Rdkafka::RdkafkaError => e
  raise MessageDeliveryError.new(e, handle)
end
```

dasch (changes requested):
Looks good so far! I think we need instrumentation for _all_ the producer methods, if nothing else then just to count the number of messages being produced.

lib/racecar/producer.rb
pfemo: Thanks - I will add that and use it in `produce_sync` as well since `rdkafka-ruby`'s `wait` is using the same values as [us per default](https://github.com/appsignal/rdkafka-ruby/blob/c02f217a189b71e3e253fe4ad59286b5fc9d4034/lib/rdkafka/abstract_handle.rb#L45-L63):

```
def wait(max_wait_timeout: 60, wait_timeout: 0.1)
      timeout = if max_wait_timeout
                  CURRENT_TIME.call + max_wait_timeout
                else
                  nil
                end
      loop do
        if pending?
          if timeout && timeout <= CURRENT_TIME.call
            raise WaitTimeoutError.new("Waiting for #{operation_name} timed out after #{max_wait_timeout} seconds")
          end
          sleep wait_timeout
        elsif self[:response] != 0
          raise_error
        else
          return create_result
        end
      end
    end
```

pfemo: We can create a seperate PR for allowing users to set the values to whatever they want.

lib/racecar.rb
dasch: Still missing this.

leonmaia (approved):
❤️ Great job! I'll release a beta version really soon with this 👍.

NobodysNightmare:
Has the corresponding beta already landed?

I also noticed that this addition is yet missing from the CHANGELOG, assuming it would belong to the "Unreleased" section.

Looking forward to this :)

PR #315, opened 2023-01-25
WIP: Thread based concurrency

bestie:
Concurrently runs a number of 'Racecar::Runner' instances in a fixed size thread pool.

Each thread starts a single `Racecar::Runner` with `Racecar::Consumer` class instance. All threads run the same consumer class, have the same config and consume partitions from the same topic(s).

`ThreadPoolRunnerProxy` can be combined with `ParallelRunner`, to run forks and threads. `ParallelRunner` is not used or battle tested (at Zendesk) and so this is still not a recommended thing to do.

Other inclusions:

- Signal handling has been moved up one level to the CLI
- Runner-like object interface standardized to `#run` #stop` `#running?`
- Test can be run locally with Docker, `export LOCAL=1`
- Some test bugs have been fixed, connections now always close and orphaned processes raise an exception


lib/racecar/config.rb
```diff
@@ -223,7 +236,8 @@ def load_consumer_class(consumer_class)
         consumer_class.name.gsub(/[a-z][A-Z]/) { |str| "#{str[0]}-#{str[1]}" }.downcase,
       ].compact.join
 
-      self.parallel_workers = consumer_class.parallel_workers
+      self.forks = consumer_class.forks
```
dasch, line 239: I think we should avoid breaking changes unless there's a _really_ good reason.

lib/racecar/config.rb
```diff
@@ -223,7 +236,8 @@ def load_consumer_class(consumer_class)
...
+      self.threads = consumer_class.threads
```
dasch, line 240: I think this should be configurable with the config itself as well; could we add a new config key and have this be e.g. `self.threads = consumer_class.threads || self.threads`?

lib/racecar/consumer.rb
```diff
@@ -7,9 +7,25 @@ class Consumer
...
+
+      # For deprecation
+      def parallel_workers
+        self.forks
+      end
+
+      # For deprecation
+      def parallel_workers=(forks)
+        self.forks = forks
+      end
```
dasch, line 28: Ah, I see there's backwards compatibility; all good!

bestie:
My pleasure!

That sounds a little more complex but totally doable. This is very much MVP concurrency, changing as little as existing code as possible.

Unless someone already has that work in progress, I can give it a try.

Is this enough an improvement to consider merging? If so, I think we could iterate towards your proposed design without breaking the API. It should also be marked experimental anyway.

dasch:
I'm not sure it provides enough value as it stands – with copy-on-write, the forking approach seems better if we're using separate consumer instances anyway.

I think the shared-main-loop approach should be pretty doable; if you look into Runner, you can see how batches are being processed sequentially – we would change that to an async model, with a fixed partition number -> worker thread mapping probably.

lib/racecar/config.rb
bestie: Agreed, it's actually aliased which you see further down the PR 👇

bestie:
79sthj

PR #319, opened 2023-03-07
Event based Kubernetes liveness probe

bestie:
This PR is functionally similar to @jbdietrich's PR #308 

The main loop of the runner touches a file and a script will exit 0 or 1 depending on whether it was touched recently enough.

Why this implementation?

- The health check is triggered by the in-built instrumentation events so doesn't clutter up the main loop code and is entirely optional.
- The 'probe' is a first class object, allowing for extensibility such as monitoring individual partition processing or rebalancing (#309)
- The script is written in Ruby, I think we agree we like more than bash and can spare the extra ~10ms of CPU to run it.
- There's some appealing refactoring and test fixes bundled with it.

This has been back-ported into Voice and will be deployed shortly. In testing it has achieved very smooth rolling deploys with minimal disruption to message processing.

After production deployment I'd be happy to update the k8s deployment section of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,6 +508,63 @@ The important part is the `strategy.type` value, which tells Kubernetes how to u
 
 Instead, the `Recreate` update strategy should be used. It completely tears down the existing containers before starting all of the new containers simultaneously, allowing for a single synchronization stage and a much faster, more stable deployment update.
 
+#### Liveness Probe
+
+Racecar comes with a built-in liveness probe, primarily for use with Kubernetes, but useful for any deployment environment where you can periodically run a process to check the health of your consumer.
+
+To use this feature:
+- set the `liveness_probe_enabled` config option to true.
+- configure your Kubernetes deployment to run `$ racecarctl liveness_probe`
+
+
+When enabled (see config) Racecar will touch the file at `liveness_probe_file_path` each time it finishes polling Kafka and processing the messages in the batch (if any).
+
+The modified time of this file can be observed to determine when the consumer last exhibited 'liveness'.
+
+Running `racecarctl liveness_probe` will return a successful exit status if the last 'liveness' event happened within an acceptable time, `liveness_probe_max_interval`.
+
+`liveness_probe_max_interval` should be long enough to account for both the Kafka polling time of `max_wait_time` and the processing time of a full message batch.
+
+On receiving `SIGTERM`, Racecar will gracefully shut down and delete this file, causing the probe to fail immediately after exit.
+
+You may wish to tolerate more than one failed probe run to accommodate for environmental variance and clock changes.
+
+See the [Configuration section](https://github.com/zendesk/racecar#configuration) for the various ways the liveness probe can be configured, environment variables being one option.
+
+Here is an example Kubernetes liveness probe configuration:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+      - name: consumer
+
+        args:
+        - racecar
+        - SomeConsumer
+
+        env:
+        - name: RACECAR_LIVENESS_PROBE_ENABLED
+          value: "true"
+
+        livenessProbe:
+          exec:
+            command:
+            - racecarctl
+            - liveness_probe
+
+          # Allow up to 10 consecutive failures before terminating Pod:
+          failureThreshold: 10
+
+          # Wait 30 seconds before starting the probes:
+          initialDelaySeconds: 30
+
+          # Perform the check every 10 seconds:
+          periodSeconds: 10
+
 #### Deploying to Heroku
 
 If you run your applications in Heroku and/or use the Heroku Kafka add-on, you application will be provided with 4 ENV variables that allow connecting to the cluster: `KAFKA_URL`, `KAFKA_TRUSTED_CERT`, `KAFKA_CLIENT_CERT`, and `KAFKA_CLIENT_CERT_KEY`.
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -5,6 +5,7 @@ require "logger"
 require "fileutils"
 require "racecar/rails_config_file_loader"
 require "racecar/daemon"
+require "racecar/liveness_probe"
 
 module Racecar
   class Cli
@@ -56,6 +57,11 @@ module Racecar
         daemonize!
       else
         $stderr.puts "=> Ctrl-C to shutdown consumer"
+      end
+
+      if config.liveness_probe_enabled
+        $stderr.puts "=> Liveness probe enabled"
+        config.install_liveness_probe
       end
 
       processor = consumer_class.new
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
+require "tmpdir"
+
 require "king_konf"
+
+require "racecar/liveness_probe"
+require "racecar/instrumenter"
 
 module Racecar
   class Config < KingKonf::Config
@@ -167,6 +172,15 @@ module Racecar
           for backward compatibility, however this can be quite memory intensive"
     integer :statistics_interval, default: 1
 
+    desc "Whether to enable liveness probe behavior (touch the file)"
+    boolean :liveness_probe_enabled, default: false
+
+    desc "Path to a file Racecar will touch to show liveness"
+    string :liveness_probe_file_path, default: "#{Dir.tmpdir}/racecar-liveness"
+
+    desc "Used only by the liveness probe: Max time (in seconds) between liveness events before the process is considered not healthy"
+    integer :liveness_probe_max_interval, default: 5
+
     # The error handler must be set directly on the object.
     attr_reader :error_handler
 
@@ -248,6 +262,35 @@ module Racecar
       end.to_h
       producer_config.merge!(rdkafka_security_config)
       producer_config
+    end
+
+    def instrumenter
+      @instrumenter ||= begin
+        default_payload = { client_id: client_id, group_id: group_id }
+
+        if defined?(ActiveSupport::Notifications)
+          # ActiveSupport needs `concurrent-ruby` but doesn't `require` it.
+          require 'concurrent/utility/monotonic_time'
+          Instrumenter.new(backend: ActiveSupport::Notifications, default_payload: default_payload)
+        else
+          logger.warn "ActiveSupport::Notifications not available, instrumentation is disabled"
+          NullInstrumenter
+        end
+      end
+    end
+    attr_writer :instrumenter
+
+    def install_liveness_probe
+      liveness_probe.tap(&:install)
+    end
+
+    def liveness_probe
+      require "active_support/notifications"
+      @liveness_probe ||= LivenessProbe.new(
+        ActiveSupport::Notifications,
+        liveness_probe_file_path,
+        liveness_probe_max_interval
+      )
     end
 
     private
--- a/lib/racecar/liveness_probe.rb
+++ b/lib/racecar/liveness_probe.rb
@@ -1,0 +1,78 @@
+require "fileutils"
+
+module Racecar
+  class LivenessProbe
+    def initialize(message_bus, file_path, max_interval)
+      @message_bus = message_bus
+      @file_path = file_path
+      @max_interval = max_interval
+      @subscribers = []
+    end
+
+    attr_reader :message_bus, :file_path, :max_interval, :subscribers
+    private     :message_bus, :file_path, :max_interval, :subscribers
+
+    def check_liveness_within_interval!
+      unless liveness_event_within_interval?
+        $stderr.puts "Racecar healthcheck failed: No liveness within interval #{max_interval}s. Last liveness at #{last_liveness_event_at}, #{elapsed_since_liveness_event} seconds ago."
+        Process.exit(1)
+      end
+    end
+
+    def liveness_event_within_interval?
+      elapsed_since_liveness_event < max_interval
+    rescue Errno::ENOENT
+      $stderr.puts "Racecar healthcheck failed: Liveness file not found `#{file_path}`"
+      Process.exit(1)
+    end
+
+    def install
+      unless file_path && file_writeable?
+        raise(
+          "Liveness probe configuration error: `liveness_probe_file_path` must be set to a writable file path.\n" \
+            "  Set `RACECAR_LIVENESS_PROBE_FILE_PATH` and `RACECAR_LIVENESS_MAX_INTERVAL` environment variables."
+        )
+      end
+
+      subscribers << message_bus.subscribe("start_main_loop.racecar") do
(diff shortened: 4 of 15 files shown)


bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
...
+file_path_and_max_interval_set? or exit(1)
+liveness_file_exists? or exit(1)
+liveness_within_timeout_interval? or exit(1)
+
+# Didn't exit so all is well
+$stderr.puts("Racecar healthcheck OK #{elapased_since_liveness_event}")
+exit(0)
+
+
+BEGIN {
```
dasch, line 15: Would moving the method definitions to the top achieve the same result as having `BEGIN` here? I'm not even sure what it does...

bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
...
+  def liveness_within_timeout_interval?
+    result = elapased_since_liveness_event < $liveness_max_interval
```
dasch, line 17:
```suggestion
    result = elapsed_since_liveness_event < $liveness_max_interval
```

bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
...
+    if result
+      true
+    else
+      $stdout.puts(
+        "Racecar healthcheck failed: No liveness within interval #{$liveness_max_interval}s. Last liveness at #{last_liveness_event_at}, #{elapased_since_liveness_event} seconds ago."
```
dasch, line 22:
```suggestion
        "Racecar healthcheck failed: No liveness within interval #{$liveness_max_interval}s. Last liveness at #{last_liveness_event_at}, #{elapsed_since_liveness_event} seconds ago."
```

bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
...
+      true
+    else
+      $stdout.puts(
+        "Racecar healthcheck failed: Bad values RACECAR_LIVENESS_PROBE_FILE_PATH=`#{$file_path}` RACECAR_LIVENESS_PROBE_MAX_INTERVAL=`#{$liveness_max_interval}`"
+      )
+      false
+    end
+  end
+
+  def elapased_since_liveness_event
```
dasch, line 49:
```suggestion
  def elapsed_since_liveness_event
```

dasch (changes requested):
Great stuff! Couple of comments.

I think it would be best to use the existing Config system for this rather than introduce new ENV vars.

Also... should `liveness_probe` either 1) be a separate thing in `exe/`, e.g. `racecar-liveness-probe` or 2) be a new subcommand in `racecarctl`?

dasch:
You should be able to extend https://github.com/zendesk/racecar/blob/master/lib/racecar/ctl.rb with a new subcommand. Perhaps pull the probe logic into a new class, e.g. `Racecar::LivenessProbe`?

bin/liveness_probe
bestie: Yes, `BEGIN` changes the flow of the file so you can put the method calls at the top. Allows the reader to see the 'what' first without the 'how' getting in the way, but I'm not wedded to it.

bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env -S ruby --disable=gems
+
+$file_path = ENV.fetch("RACECAR_LIVENESS_PROBE_FILE_PATH", "")
+$liveness_max_interval = Integer(ENV.fetch("RACECAR_LIVENESS_PROBE_MAX_INTERVAL")) rescue nil
```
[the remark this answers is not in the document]
bestie: I don't have a strong opinion about that, my thinking here was that I was competing with a bash script.

Something that executes quickly with no dependencies is more palatable to ops people and they will likely want to put those values in the `ENV` anyway.

If you don't buy that argument then I'm happy to change it.

bin/liveness_probe
bestie: I wish I could spell :)

bin/liveness_probe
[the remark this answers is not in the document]
bestie: I would also be concerned that people would assume their Rails app is being loaded for the probe, that wouldn't make sense to do that but someone will assume that.

I will of course add docs and we can try to make that clear.

bin/liveness_probe
```diff
@@ -0,0 +1,56 @@
...
+
+file_path_and_max_interval_set? or exit(1)
+liveness_file_exists? or exit(1)
+liveness_within_timeout_interval? or exit(1)
```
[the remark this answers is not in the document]
bestie: I actually did that originally and then thought this might be easier for ops / non Rubyists who would understand `exit(1)` immediately.

lib/racecar/runner.rb
```diff
@@ -67,6 +67,8 @@ def run
       loop do
         break if @stop_requested
         resume_paused_partitions
+
+        @instrumenter.instrument("start_main_loop", instrumentation_payload)
```
bestie, line 71: @dasch What do you think about this? I added it because it was more convenient for me but there are some alternatives.

- Add something like a 'starting' event and use that with together with 'main_loop'
- Use a listener object with `start` method to trigger at the start of the block rather than at the end

bestie:
> Great stuff! Couple of comments.
> 
> I think it would be best to use the existing Config system for this rather than introduce new ENV vars.
> 
> Also... should `liveness_probe` either 1) be a separate thing in `exe/`, e.g. `racecar-liveness-probe` or 2) be a new subcommand in `racecarctl`?

Definitely a neater solution but let me know what you think about the "easier for ops people" argument.

bin/liveness_probe
dasch: Back in the old days I'd just use a `main` method that I would call from the bottom of the file. But I think probably we'll want to call into this code from `racecarctl` anyway.

lib/racecar/runner.rb
dasch: I actually _think_ you can subscribe to the start and finish of instrumented blocks, [but it might be more trouble than it's worth](https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activesupport/lib/active_support/notifications/fanout.rb#L129). Basically, instead of passing a block or callable to `subscribe` you'd pass an object that responds to both `start` and `finish`.

Probably your solution is simpler :D

bin/liveness_probe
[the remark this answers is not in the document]
dasch: I think we should target the _dev_ ops folks, basically the engineers who write the consumers, and they'd be surprised by having a separate config system. Many people like having their config in a single YAML file.

bin/liveness_probe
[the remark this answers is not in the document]
dasch: 🤷 I don't think folks will necessarily read through the code that much, so long as the behavior is documented. I'd rather stick with Rubyisms.

dasch:
@leonmaia would you also like to review this?

deepredsky (commented):
We actually wanted to rewrite the bash part in https://github.com/zendesk/racecar/pull/308 into ruby as well, so thanks for this PR! :)

bin/liveness_probe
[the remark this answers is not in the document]
deepredsky: We had some discussions around the this when mobbing on https://github.com/zendesk/racecar/pull/308. With the bash version in that PR, reusing the config was not easy. I agree with @dasch about using the same config although it comes with additional load. cc @jbdietrich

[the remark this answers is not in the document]
jbdietrich: FWIW, I would tend to agree that it is better to make this consistent with the existing config API.

bin/liveness_probe
bestie: Oh same, but then I got old and weird 😆

lib/racecar/runner.rb
bestie: Yes, that what I was saying, you have to pass an object that responds to `#start` and `#finish` which is annoying extra step.

I think the fact it wasn't obvious to either of us means we should keep the start events.

lib/racecar/cli.rb
```diff
@@ -58,6 +59,11 @@ def run
         $stderr.puts "=> Ctrl-C to shutdown consumer"
       end
 
+      if config.liveness_probe_file_path
+        $stderr.puts "=> Liveness probe enabled"
+        config.install_liveness_probe
+      end
```
dasch, line 65: I know this review has dragged out a bit, but there's one last change I think we should make 😄 

I think it would make sense to have a separate `liveness_probe_enabled` boolean config flag, and to auto-generate the file path if not explicitly set, using [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html). Since we don't need to synchronize the path between the Kubernetes probe config and Racecar now that the latter has an abstraction in place, it's probably better to just auto-generate tempfiles for the liveness data, and not need users to worry about it.

lib/racecar/config.rb
```diff
@@ -250,6 +256,35 @@ def rdkafka_producer
...
+    attr_writer :instrumenter
+
+    def install_liveness_probe
+      liveness_probe.tap(&:install)
+    end
+
+    def liveness_probe
+      require "active_support/notifications"
+      @liveness_probe ||= LivenessProbe.new(
+        ActiveSupport::Notifications,
```
dasch, line 282: Should this be `instrumenter`?

lib/racecar/liveness_probe.rb
```diff
@@ -0,0 +1,73 @@
...
+      @max_interval = max_interval
+      @subscribers = []
+    end
+
+    attr_reader :message_bus, :file_path, :max_interval, :subscribers
+    private     :message_bus, :file_path, :max_interval, :subscribers
+
+    def check_liveness_within_interval!
+      liveness_event_within_interval? or
+        raise "Racecar healthcheck failed: No liveness within interval #{max_interval}s. Last liveness at #{last_liveness_event_at}, #{elapsed_since_liveness_event} seconds ago."
```
dasch, line 17: Could we just write to stderr and `exit(1)` here? Otherwise we'll clobber the logs with stack traces.

lib/racecar/liveness_probe.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def uninstall
+      subscribers.each { |s| message_bus.unsubscribe(s) }
+    end
+
+    private
+
+    def elapsed_since_liveness_event
+      Time.now - last_liveness_event_at
```
dasch, line 50: There _is_ an error scenario here where the local clock jumps. Should we consider using `Process.clock_gettime(Process::CLOCK_MONOTONIC)` and storing the values in the files instead of relying on `mtime` and `Time.now`? Considering that a clock jump might actually cause the kpods to be terminated.

bestie: I always have the same thought when I see `Time.now`!

In the case of forward clock jump a single check could fail, but the next loop will will correct the situation.

I don't think killing a process on one liveness check failure is normal so I didn't add the complexity. If we want to support that use-case then np.

lib/racecar/liveness_probe.rb
bestie: Yes, that's a good call.

Generally I prefer to raise, but the stack trace doesn't help.

lib/racecar/config.rb
bestie: `LivenessProbe` needs to subscribe to the instrumentation events and the `Racecar::Instrumenter` class only implements `#instrument`.

The other reason it's not `instumenter` is that we have this optional `ActiveSupport::Notifications` dependency that is actually required for the liveness probe.

I would be super happy to have the flexibility of passing in your own instrumentation backend but I'm not sure anyone wants that.

lib/racecar/liveness_probe.rb
dasch: You're right, it's probably not something that could trigger a Pod termination. We can always add it in later on :+1:

lib/racecar/cli.rb
dasch: Looks good! And yes, you're right – I forgot we had multiple processes in the mix here 🤦

dasch: The default path should probably be `"#{Dir.tmpdir}/racecar-liveness` since we want to use the proper system TMPDIR. Otherwise we might get into trouble on e.g. kpods that use read-only file systems for most paths.

bestie: Yes! I was looking through the docs for `Dir.tmpdir`! Couldn't remember how to get that.

Adding now.

bestie: You have to `require "tmpdir"` to get the method. Very well hidden!

dasch: Sneaky!

dasch (approved):
Looks good! :rocket:

README.md
```diff
@@ -508,6 +508,29 @@ The important part is the `strategy.type` value, which tells Kubernetes how to u
...
+Running `racecarctl liveness_probe` will return a successful exit status if the last 'liveness' event happened within an acceptable time, `liveness_probe_max_interval`.
+
+`liveness_probe_max_interval` should be long enough to account for your the longest possible Kafka polling time of `max_wait_time` and the processing time of a full message batch.
+
+On receiving `SIGTERM` Racecar will gracefully shut down and delete this file, causing the probe to fail immediately after exit.
+
+You may wish to tolerate more than one failed probe run to accommodate for environmental variance and clock changes.
+
+See the Config section for the various ways liveness probe settings can be configured, environment variables being one option.
+
```
dasch, line 533:
```suggestion
Here is an example Kubernetes liveness probe configuration:

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      containers:
      - name: consumer

        args:
        - racecar
        - SomeConsumer

        env:
        - name: RACECAR_LIVENESS_PROBE_ENABLED
          value: "true"

        livenessProbe:
          exec:
            command:
            - racecarctl
            - liveness_probe

          # Allow up to 10 consecutive failures before terminating Pod:
          failureThreshold: 10

          # Wait 30 seconds before starting the probes:
          initialDelaySeconds: 30

          # Perform the check every 10 seconds:
          periodSeconds: 10
```

```

**Review by deepredsky — APPROVED** [review]

File: README.md [hunk 13]
```diff
@@ -530,7 +530,39 @@ On receiving `SIGTERM` Racecar will gracefully shut down and delete this file, c
...
+          # Allow up to 10 consecutive failures before terminating Pod:
+          failureThreshold: 10
+
+          # Wait 30 seconds before starting the probes:
+          initialDelaySeconds: 30
+
+          # Perform the check every 10 seconds:
+          periodSeconds: 10
 #### Deploying to Heroku
```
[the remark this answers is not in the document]
bestie: 🤦

dasch:
Alrighty! Let's add an entry to the CHANGELOG and I can merge this!

leonmaia (approved):

PR #320, opened 2023-03-09
use proper onomatopoeia when booting Racecar

Drew-Goddyn:
When a consumer process is initiated, we communicate to developers that Racecar is booting through a cute onomatopoeia.

Its not clear what model of car the developers of Racecar have (perhaps electric or some sort of scandinavian model), but according to wikipedia, the correct sound a car makes is "vrooom" not "wrooom": https://en.wikipedia.org/wiki/Vroom.

This PR modifies the sound to not break immersion for developers.

### DIFF
--- a/lib/racecar/cli.rb
+++ b/lib/racecar/cli.rb
@@ -50,7 +50,7 @@ module Racecar
         configure_datadog
       end
 
-      $stderr.puts "=> Wrooooom!"
+      $stderr.puts "=> Vrooooom!"
 
       if config.daemonize
         daemonize!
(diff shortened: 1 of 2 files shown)


leonmaia:
Thanks for the PR @Drew-Goddyn 😆. It seems some tests in the `cli_spec.rb` file are broken. Can you please update them?
https://github.com/zendesk/racecar/blob/61f489c02abeac56e228813e57093922a6545c56/spec/cli_spec.rb#L16-L20

Drew-Goddyn:
> Based on the list in the Wikipedia, I guess `Bruumm` would be the most common between different languages. Unfortunately we cannot cover it all, maybe in a new iteration we make it dependent on locale of the OS. Screenshot 2023-03-15 at 11 00 37 AM

You're right, I was overly american-centric

PR #331, opened 2023-04-14
Integration test performance and refactoring

bestie:
TL;DR Integration specs 3x faster ~45s -> 15s

This is a huge itch-scratch PR that addresses _most_ of the issues I have with the integration specs. Hopefully we're in agreement and maybe it's an easy merge 🤞🤷😬 🗒️ 

Here's the full laundry list:

- Replace Kafka CLI with `librdkadfka` admin connection for creation and
  deletion of test topics.
- Racecar::CLI implements `#stop` which is now used instead of the test
  process sending `SIGINT` to itself.
- Hook into rebalance callbacks to get a fast, reliable indication of assignments for test setup, again replacing slow CLI tools.
- Use a pipe in forked consumer processes to signal assignments back to
  the test.
- Tag messages with PIDs to confirm concurrency via message state.
- Reset implicitly installed signal handlers between tests.
- Reset Racecar config between tests.
- Use a consistent test consumer class for integration tests, rebuild it
  for every test.
- Ensure _all_ child processes have shutdown, print a warning after
  timeout.
- Removed `#in_background` and callback mechanism in favor of running
  the consumer in a thread and controlling the test from the main
  thread.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -14,6 +14,7 @@ GEM
       minitest (>= 5.1)
       tzinfo (~> 2.0)
       zeitwerk (~> 2.3)
+    byebug (11.1.3)
     coderay (1.1.3)
     concurrent-ruby (1.2.2)
     diff-lcs (1.5.0)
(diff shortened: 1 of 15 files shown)


leonmaia (approved):
Great work, LGTM 👍

PR #339, opened 2023-05-31
Add support for cooperative-sticky assignment strategy

bestie:
Features:
- Support for cooperative-sticky assignment strategy
- Consumer class partition assign and revoke callbacks

Tests:
- Refactor to share a common consumer

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     racecar (2.9.0.beta1)
       king_konf (~> 1.0.0)
-      rdkafka (~> 0.12.0)
+      rdkafka (~> 0.13.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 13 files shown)


deepredsky:
The changes in this PR looks good. But there is reported issue about cooperative sticky failing with a  crash on close prior to first rebalance.

Deep dive and some suggestion fix here: https://github.com/confluentinc/librdkafka/issues/4312

mensfeld:
> Deep dive and some suggestion fix here

Just a reminder that my bypass code is released under **LGPL** and not MIT (talking about karafka, not rdkafka-ruby or waterdrop).

mensfeld:
ah and this is a bypass. 2.2.1 librdkafka should be the one to include the correct fix and then it should be "passed" on to rdkafka-ruby via a dep bump.

deepredsky (approved):

mensfeld:
@deepredsky just a heads up: 2.3.0 has been released. I should ship 0.14.0 rdkafka-ruby in the upcoming weeks once I test it fully. All should be good then.

PR #340, opened 2023-06-22
Fix: Runner adds exceptions to instrumentation payload

bestie:
The DataDog integration checks for `:exception` in the instrumentation payload but `Runner` does not add it.

Looks like an oversight to me. Should it be added?

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -201,6 +201,7 @@ module Racecar
             consumer.store_offset(message)
           end
         rescue => e
+          instrumentation_payload[:exception] = e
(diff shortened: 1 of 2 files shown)


bestie:
Example of an exception check https://github.com/zendesk/racecar/blob/b1b05740eb95190460af53d09d4195f99a7de3e1/lib/racecar/datadog.rb#L101

deepredsky (commented):
there is no need to set the exception manually, we are using [active supports notification mechanism](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) that does this automatically for exceptions.

PR #350, opened 2023-09-01
Send exceptions to `process_batch` instrumenter

lairen:
When exceptions are raised by the processing code in `process_batch`, the `with_pause` method is swallowing them, so the instrumentation does not receive the exception

This PR makes it so that the instrumentation receives the processing exceptions.


deepredsky (approved):

deepredsky:
@lairen would you be able to add a test to cover this?

lairen:
@deepredsky done. i also added the specs for the `process_message` instrumentation

PR #355, opened 2023-09-29
Fix config load for liveness probe

deepredsky:
Some rails apps might chose to not use the racecal.yml file. Let's load known
config files if they are present similar to how cli handles them

### DIFF
--- a/lib/racecar/ctl.rb
+++ b/lib/racecar/ctl.rb
@@ -36,8 +36,10 @@ module Racecar
       require "racecar/liveness_probe"
       parse_options!(args)
 
-      if ENV["RAILS_ENV"]
-        Racecar.config.load_file("config/racecar.yml", ENV["RAILS_ENV"])
+      RailsConfigFileLoader.load! unless config.without_rails?
+
+      if File.exist?("config/racecar.rb")
+        require "./config/racecar"
       end
 
       Racecar.config.liveness_probe.check_liveness_within_interval!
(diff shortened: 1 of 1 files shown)


lib/racecar/ctl.rb
```diff
@@ -36,10 +36,14 @@ def liveness_probe(args)
       require "racecar/liveness_probe"
       parse_options!(args)
 
-      if ENV["RAILS_ENV"]
+      if ENV["RAILS_ENV"] && File.exist?("config/racecar.yml")
         Racecar.config.load_file("config/racecar.yml", ENV["RAILS_ENV"])
```
dasch, line 40: Perhaps we should use [this method](https://github.com/zendesk/racecar/blob/master/lib/racecar/rails_config_file_loader.rb#L4) for loading the config?

deepredsky: @dasch I've updated it to behave exactly as the cli load them

dasch (approved):

dasch:
Thanks!

lib/racecar/ctl.rb
haberbyte: I am wondering if loading the config like this (`RailsConfigFileLoader`) is the right thing here, doesn't that mean the liveness probe will always try to boot the entire Rails app just to check the modified time of a file?
There is the `without_rails` config option, but what if the consumer needs Rails, then the liveness probe will do all that extra work for nothing.

Issue #356, opened 2023-10-10
ActiveSupport::Logger.broadcast dropped in Rails 7.1


zubin:
After upgrading to Rails 7.1, starting a consumers fails with the following:

```
=> Detected Rails, booting application...
=> Crashed: NoMethodError: undefined method `broadcast' for ActiveSupport::Logger:Class
[...]racecar-2.9.0/lib/racecar/rails_config_file_loader.rb:30:in `load!'
```

Is a fix in the pipeline? If not I'll make a PR shortly.

gkosmo:
I'm curious as to why this PR was closed ( I would have liked that fix :grin:  )

PR #357, opened 2023-10-10
Resolve Rails 7.1 logger incompatibility

zubin:
Resolves https://github.com/zendesk/racecar/issues/356

There don't appear to be any tests which cover this but I've verified that it works locally.

### DIFF
--- a/lib/racecar/rails_config_file_loader.rb
+++ b/lib/racecar/rails_config_file_loader.rb
@@ -26,7 +26,11 @@ module Racecar
           console = ActiveSupport::Logger.new($stdout)
           console.formatter = Rails.logger.formatter
           console.level = Rails.logger.level
-          Rails.logger.extend(ActiveSupport::Logger.broadcast(console))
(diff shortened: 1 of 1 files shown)


gkosmo:
Out of curiosity, why was this PR closed ? This fix looks good to me ( it's the same strategy as used on sidekiq )

deepredsky:
The patch here should work, @zubin I m happy to merge this patch unless you found some issue with this change

zubin:
> The patch here should work, @zubin I m happy to merge this patch unless you found some issue with this change

No issues, it's been working fine. 👍

dorianmarie:
@deepredsky Could you release a new version please?

dorianmarie:
Oh I see there is 2.10.0.beta2, why not a non-beta version?

deepredsky:
> Oh I see there is 2.10.0.beta2, why not a non-beta version?

2.10.0 is released now

PR #363, opened 2024-01-18
Configurable strategy for consuming multiple topics

bestie:
The current ConsumerSet implementation will only switch topics when the current topic returns no messages after the poll timeout.

This is not practical when one or more of the topics has a 'high' message volume as the broker will always return a message and Racecar may never switch to poll the other topics.

* 'high' here means message frequency (hz) > poll interval (s).


dasch:
Any reason to make this configurable? Round robin seems like the only setup that makes sense here...

PR #364, opened 2024-02-05
Improved forking implementation

bestie:
* Set config value `forks` instead of `parallel_workers`
* Supervisor process forks worker `n` processes
* Supervisor terminates workers when it is stopped
* Workers monitor supervisor and terminate themselves if supervisor exits to prevent zombie processes
* Supervisor runs pre-fork hook before forking
* Workers run post-fork hook immediately after forking


dasch:
Would this _replace_ the existing parallel workers setup? If not, why have two?

bestie:
I think it should replace it but for now I went with the lowest friction option of complete separate feature.

There are differences in behavior on error and when processes die unexpectedly, so I was imagining we could deprecate current 'parallel workers' implementation and remove it in 3.x. If we're up for messing with an 'experimental' in a minor release I do not object.

dasch:
Probably up to the owning team, but it should at least be documented that they're two implementations of the same feature, one of them experimental (for now).

lib/racecar/forking_runner.rb
```diff
@@ -0,0 +1,131 @@
...
+        rescue Errno::ESRCH
+        end
+      end
+    end
+
+    def check_workers
+      pids.each do |pid|
+        unless worker_running?(pid)
+          $stdout.puts("A forker worker has exited. Shuttin' it down ...")
+          stop
```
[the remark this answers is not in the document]
bestie: Yes, it would unexpected for this to happen, so rather than try to recover from a potentially bad state the whole thing shuts down so Kubernetes can bring up a new replica.

I'm not against trying to recover and replace a lost worker process but that is quite a bit more complex.

spec/integration/forking_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+      start_racecar
+
+      wait_for_assignments(forks)
+      publish_messages
+      wait_for_messages
+
+      Process.kill("KILL", worker_pids[0])
+      Process.waitpid(supervisor_pid)
+
+      expect_processes_to_have_exited(worker_pids)
```
HeyNonster, line 95: If we want to k8s to spin up a new replica shouldn't this also kill the `supervisor_pid`?

lib/racecar/forking_runner.rb
```diff
@@ -0,0 +1,131 @@
...
+      liveness_monitor.parent_post_fork
+
+      install_signal_handlers
+
+      wait_for_child_processes
+    end
+
+    def stop
+      @running = false
+      $stdout.puts "Racecar::ForkingRunner runner stopping #{Process.pid}"
```
HeyNonster, line 47: Should the supervisor pid be terminated too after each worker is killed?

HeyNonster: Actually, maybe that happens automatically 🤔

PR #369, opened 2024-03-27
Avoid loading config files when executing the liveness probe check

bestie:
## Proposed change

Add a new configuration option that prevents the liveness probe command from loading config files.

## Why?

It's possible, even common to end up Rails specific logic Racecar config files.

`Racecar.config.logger = Rails.logger` for example.

In these cases it's not desirable for the liveness probe to use the same configuration mechanism. A new option,`RACECAR_LIVENESS_PROBE_SKIP_CONFIG_FILES` ensures this doesn't happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Allow the liveness probe command to skip loading config files
+
 ## 2.11.0.beta3
 
 * Fix bug with domain socket support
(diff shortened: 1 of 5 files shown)


deepredsky (approved):
I think we need a note in the readme about this. Also can you add an entry in changelog 🙏

bestie:
@deepredsky I've changed approach on this a little, updated the readme and added to the changelog.

deepredsky (approved):

angelim (approved):

PR #365, opened 2024-02-28
feat: support uds in datadog statsd

andrewthauer:
This adds support for unix domain sockets with datadog statsd

PR #370, opened 2024-04-03
Fix datadog unix socket path

ewhorton:
## What

The latest beta release broke the consumer due to exception:
`NoMethodError:  undefined method 'socket_path'`

This PR should fix this


ewhorton:
@deepredsky @andrewthauer this should fix a bug with [this feature](https://github.com/zendesk/racecar/pull/365)

I've tested locally and it resolved the issue

JaydaKad (approved):

PR #371, opened 2024-05-07
use karafka-core instead of rdkafka

godwhoa:


godwhoa:
never mind this, opened a PR on the wrong repo 😬

PR #395, opened 2025-02-26
Allow processing partitions concurrently

dasch:
Use concurrent-ruby to process each partition in a fetch response concurrently. This maintains in-order processing within partitions but allows using multiple cores to process messages in *different* partitions at the same time.


henrikbjorn:
This would be awesome. Ready to test, or is something needed to bring it to completion ?

PR #410, opened 2026-02-16
Add thread-based message processing

ChrisGadek1:
This PR adds an optional opportunity to run a thread pool and allow threads to consume and process messages (and message batches) in parallel. Key changes:
- Added an optional parameter `self.parallel_batches_executors` in the Consumer class (accordingly to README.md). If this parameter is set and is valid, then the thread pool will be launched. If not, the default processing will be used.
- Added `thread_safe` method in the Consumer to wrap operations, which require thread-safeness
- Updated tests

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,4 +81,4 @@ DEPENDENCIES
   timecop
 
 BUNDLED WITH
-   2.6.2
+   2.4.19
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,4 +81,4 @@ DEPENDENCIES
   timecop
 
 BUNDLED WITH
-   2.4.19
+   2.6.2
--- a/lib/racecar/config.rb
+++ b/lib/racecar/config.rb
@@ -240,6 +240,10 @@ module Racecar
       if max_pause_timeout && !pause_with_exponential_backoff?
         raise ConfigError, "`max_pause_timeout` only makes sense when `pause_with_exponential_backoff` is enabled"
       end
+
+      if shutdown_timeout <= 1
+        raise ConfigError, "`shutdown_timeout` must be greater than 0"
+      end
     end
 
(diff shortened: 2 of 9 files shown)


lib/racecar/runner.rb
```diff
@@ -46,6 +48,15 @@ def setup_pauses
       }
     end
 
+    def setup_multithreading
+      if !processor.parallel_batches_executors.nil? && processor.parallel_batches_executors > 0
```
razumau: This `if` is a bit difficult to read: would you like to extract these two into something like `multithreading_enabled?`?

Gemfile.lock
```diff
@@ -81,4 +81,4 @@ DEPENDENCIES
   timecop
 
 BUNDLED WITH
-   2.6.2
+   2.4.19
```
razumau: Seems to be committed by mistake.

lib/racecar/runner.rb
ChrisGadek1: Right, extracted it to a separate method

Gemfile.lock
ChrisGadek1: Yes, removed that

lib/racecar/config.rb
```diff
@@ -237,6 +240,10 @@ def validate!
       if max_pause_timeout && !pause_with_exponential_backoff?
         raise ConfigError, "`max_pause_timeout` only makes sense when `pause_with_exponential_backoff` is enabled"
       end
+
+      if shutdown_timeout <= 1
+        raise ConfigError, "`shutdown_timeout` must be greater than 0"
```
dasch, line 245:
```suggestion
      if shutdown_timeout <= 0
        raise ConfigError, "`shutdown_timeout` must be greater than 0"
```

Folks might want 0.5?

PR #413, opened 2026-03-17
Introduction of multithreading in Racecar

ChrisGadek1:
## Add per-partition multithreaded processing

  Introduces opt-in multithreaded message processing where each topic-partition gets a
  dedicated worker thread. Disabled by default — existing single-threaded behaviour is
  unchanged.
  
  ### What changed

  #### New: `AsyncPartitionProcessor`
  
  When `multithreaded_processing_enabled` is `true`, `Runner` creates one
  `AsyncPartitionProcessor` per topic-partition (lazily, on first message). Each processor:

  - Wraps a `PartitionProcessor` (the shared processing core) and a dedicated Ruby thread
    backed by a `Queue`
  - Blocks on `Queue#pop` — the thread wakes only when a message (or batch) is pushed.
    Shutdown and rebalance signal the thread by pushing a `nil` sentinel, which causes the
    loop to `break` and run its `ensure` cleanup (`teardown`)
  - Supports both `process` and `process_batch` consumers: the routing decision
    (`consumer_class.method_defined?(:process_batch)`) is resolved once at thread-spawn time
    against the class, not an instance
  - **Backpressure** — pauses the Kafka partition when `@queue.size` reaches
    `multithreaded_processing_max_queue_size`; resumes when it drains below
    `multithreaded_processing_resume_threshold × max_queue_size`. Queue depth is measured in
    `Queue` entries via `Queue#size`. The backpressure flag is a `Concurrent::AtomicBoolean`
    for cross-thread visibility, and pause/resume act on a per-partition `TopicPartitionList`
    built at construction
  - **Rebalance / shutdown** — `rebalance!` and `shut_down!` delegate to the underlying
    `PartitionProcessor` (which flips its state flag and signals any in-flight backoff wait)
    and then push the `nil` sentinel to unblock the queue so the thread can exit

  #### New: `PartitionProcessor`

  The shared processing core used in both modes. Each `PartitionProcessor`:

  - Accepts `consumer_class_instance:` (a pre-built instance, not a class). In
    single-threaded mode all processors share one instance; in multithreaded mode each
    `AsyncPartitionProcessor` creates its own via `consumer_class.new`
  - In multithreaded mode, calls `configure` on its consumer instance at construction,
    wiring it to the shared producer, consumer set, instrumenter, and config (in
    single-threaded mode the shared instance is configured once by `Runner#run`)
  - Before each `process` / `process_batch`, reconfigures the consumer instance if its
    producer was reset by an error handler — detected via the instance's own
    `@producer&.closed?` (multithreaded mode only)
  - Owns a single `Pause` object for its topic-partition, plus a `Mutex` + `ConditionVariable`
    used to implement an interruptible error backoff

  The processor selects its error-handling strategy at runtime:

  - **Multithreaded** — retries in a loop; between attempts it waits on a `ConditionVariable`
    for `pause.backoff_interval` (interruptible — a rebalance or shutdown wakes it
    immediately). It `Thread.exit`s on rebalance and `break`s on shutdown or when
    `pause_timeout <= 0`
  - **Single-threaded** — pauses the Kafka partition on error and re-raises;
    `resume_all_paused_partitions` runs before each main-loop iteration to un-pause expired
    partitions
  
  `teardown` flushes pending delivery handles (`deliver!`) and calls the user's `teardown`
  hook; the `deliver!` is skipped when `rebalancing` is set (partition revoked mid-run).

  #### Updated: `Runner`

  `Runner` now accepts the consumer class itself (not an instance). Key changes:

  - Builds a single `consumer_class.new` instance in the constructor for `statistics_callback`
    setup and, in single-threaded mode, as the shared processing instance passed to all
    `PartitionProcessor`s (it is configured once in `run`)
  - Tracks processors in a `Concurrent::Hash` keyed by `"topic/partition"`. Processors are
    created lazily on first message:
    - **multithreaded** → one `AsyncPartitionProcessor` per partition, each building its own
      consumer instance internally
    - **single-threaded** → one `PartitionProcessor` per partition, all sharing the single
      consumer instance
  - The main loop is wrapped in `begin … ensure` so `shutdown_processors_and_wait` and
    `consumer.commit` run on both normal and exceptional exit
  - `consumer.close`, Datadog teardown, and `shut_down` instrumentation live in a separate
    outer `ensure`, each individually protected
  - `shutdown_processors_and_wait`:
    - **multithreaded** → signals every processor via `shut_down!`, then `join`s each thread
      with `multithreaded_processing_shutdown_timeout` (individual `join` failures are rescued
      so remaining threads are still joined)
    - **single-threaded** → calls `deliver!` then `teardown` once on the shared instance, with
      `teardown` in an `ensure`

  #### Shared producer on `ConsumerSet`

  The rdkafka producer moved from `Runner` to `ConsumerSet`, so it can be shared across
  per-partition threads:

  - `ConsumerSet#producer` — lazily-initialised shared producer with delivery callback,
    guarded by `@producer_mutex`
  - `ConsumerSet#reset_producer!` — closes the current producer and clears it under
    `@producer_mutex`, serialising concurrent resets from processing threads
  - `ConsumerSet#close` resets the producer alongside closing all consumers
  - Delivery handles remain per consumer-class-instance — no thread-local keying needed,
    since each multithreaded processor owns its own instance

  #### Consumer instance isolation
  
  - **Single-threaded mode**: one shared consumer instance across all partitions, matching
    the original behaviour. Instance state (counters, caches, etc.) is shared as before
  - **Multithreaded mode**: each `AsyncPartitionProcessor` builds its own instance via
    `consumer_class.new`, so user consumer code does **not** need to be thread-safe — each
    partition thread has exclusive access to its own instance

  #### Updated: `Pause`
  
  `Pause.new_from_config(config)` replaces the per-partition pause construction previously
  done inline in `Runner`. It returns a single `Pause` instance (each `PartitionProcessor`
  owns one for its topic-partition), and the `pause_timeout == 0` and `pause_timeout == -1`
  cases are merged into a single `nil`-timeout branch.

  #### Updated: `RebalanceListener`

  Now receives the `config`, instrumenter, and the shared `partition_processors` map. On
  `on_partitions_revoked`, for each revoked topic-partition it calls `processor.rebalance!`
  (which signals the worker to exit) and removes the processor from the map, so the next
  assignment creates a fresh one.

  #### Removed: `@paused_tpls` tracking in `ConsumerSet`

  The `@paused_tpls` hash (and the paused-consumer fallback in `find_consumer_by`) has been
  removed. Backpressure state is now tracked per-processor via
  `AsyncPartitionProcessor#backpressure_paused` (`Concurrent::AtomicBoolean`), and
  error-pause state via the `Pause` object on `PartitionProcessor`.

  #### New config options
  
  | Option | Default | Description |
  |---|---|---|
  | `multithreaded_processing_enabled` | `false` | Enable per-partition threads |
  | `multithreaded_processing_max_queue_size` | `1000` | Max queued batches before backpressure kicks in |
  | `multithreaded_processing_resume_threshold` | `0.5` | Resume paused partition when queue drops below this fraction of max |
  | `multithreaded_processing_shutdown_timeout` | `300` | Seconds to wait per thread during graceful shutdown |

  ---

  ### Integration tests (`spec/integration/multithreading_spec.rb`)

  - One thread spawned per partition; threads are distinct across partitions and workers
  - Multi-topic consumers each get independent per-partition threads
  - Backpressure pauses the partition when the queue fills and auto-resumes once it drains
    below half capacity
  - Rebalance exits the revoked partition's thread cleanly; remaining messages are
    re-delivered by the new owner
  - Error retry with configurable backoff eventually processes the message
  - `process_batch` consumers are supported
  - Produced messages preserve payload, key, and partition
  - Graceful shutdown drains all queued messages before exit
  - Parallel workers each spawn threads only for their assigned partitions

## Diagram schema

kafka_multithreading

## Impact on performance

The new multithreading was tested on real, production Zendesk kafka consumers for about 5 days. During that period, the overall throughput per replica has increased about 5x, which proves that multithreading can help reduce number of replicas of consumers and use computational resources more effectively for message processing.

image


lib/racecar/multi_threaded_processor.rb
```diff
@@ -0,0 +1,211 @@
...
+          processor.process_batch(msgs)
+          finalize_messages_processing(thread_key, msgs.last)
+        end
+      else
+        msg = msgs.first
+        instrument_process_message(instrumentation_payload(msg.original_message)) do
+          processor.process(msg)
+          finalize_messages_processing(thread_key, msg)
+        end
+      end
```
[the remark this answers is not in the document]
ChrisGadek1: This comment is not relevant, because according to the schema, revoked partition messages are dropped and we let the new owner of the partition to start polling messages from last committed offset. But yes, the PR description should be updated

lib/racecar/consumer_set.rb
```diff
@@ -123,7 +124,11 @@ def resume(topic, partition)
       @paused_tpls.delete(topic) if @paused_tpls[topic].empty?
     end
 
-    alias :each :each_subscribed
+    def paused?(topic, partition)
+      @paused_tpls[topic].key?(partition)
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: This helper method is only used in tests, so there's no need to make changes here

lib/racecar/threads_orchestrator.rb
```diff
@@ -0,0 +1,211 @@
...
+        spawn_thread(thread_key) do |current_thread_key, processed_messages|
+          process_messages(current_thread_key, processed_messages)
+        end
+        logger.debug "Spawned thread for topic: #{topic}, partition: #{partition}"
+      end
+      Array(messages).each { |m| @thread_queues[thread_key] << m }
+      wakeup(thread_key)
+
+      maybe_apply_backpressure(thread_key, topic, partition, messages)
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: The risk of crashes is negligible and having per-key mutexes prevent from synchronising on every metadata fetch, which is strictly related to the thread it's running. Having a single mutex would create a bottleneck on unnecessary synchronisations. Although, I've introduced a single mutex for `deliver!` and `store_offset` operations

lib/racecar/threads_orchestrator.rb
```diff
@@ -0,0 +1,211 @@
...
+      @threads[thread_key] = Thread.new do
+        Thread.current.name = "Racecar thread for #{thread_key}"
+        loop do
+          maybe_stop_or_exit(thread_key)
+          msgs = acquire_messages_from_queue(thread_key)
+          process_with_error_handling_and_retrying(thread_key, msgs) do
+            yield thread_key, msgs
+          end
+        end
+        logger.debug "Thread for #{thread_key} exiting"
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed that

lib/racecar/threads_orchestrator.rb
```diff
@@ -0,0 +1,211 @@
...
+    end
+
+    def finalize_messages_processing(thread_key, msg)
+      @thread_mutexes[thread_key].synchronize do
+        processor.deliver!
+      end
+      @thread_mutexes[thread_key].synchronize do
+        consumer.store_offset(msg)
+      end
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Added a single mutex

lib/racecar/config.rb
```diff
@@ -194,6 +194,16 @@ class Config < KingKonf::Config
...
     string :multi_subscription_strategy, allowed_values: %w(round-robin exhaust-topic), default: "exhaust-topic"
 
+    desc "Whether multithreaded processing is enabled"
+    boolean :multithreaded_processing_enabled, default: false
+
+    desc "Max size of the queue of messages waiting to be processed when multithreaded processing is enabled"
+    integer :multithreaded_processing_max_queue_size, default: 1000
+
+    desc "Whether to fetch all messages in a batch when multithreaded processing is enabled. If false, messages will be fetched one by one even if `process_batch` was used.
+                    If set to true, then the whole FIFO queue will be emptied on each processing execution, and batches passed to `process_batch` may be of variable size. Defaults to true."
```
dasch: Use the Ruby heredoc syntax here in order to avoid weird indentation on the second line.

lib/racecar/threads_orchestrator.rb
```diff
@@ -0,0 +1,215 @@
...
+      topic = messages.is_a?(Array) ? messages.first.topic : messages.topic
+      partition = messages.is_a?(Array) ? messages.first.partition : messages.partition
+      thread_key = self.class.thread_key(topic, partition)
+      unless @thread_queues[thread_key] && @threads[thread_key]&.alive?
+        spawn_thread(thread_key) do |current_thread_key, processed_messages|
+          process_messages(current_thread_key, processed_messages)
+        end
+        logger.debug "Spawned thread for topic: #{topic}, partition: #{partition}"
+      end
+      Array(messages).each { |m| @thread_queues[thread_key] << m }
```
dasch: Would it make more sense to push _batches_ of messages onto the queue instead of having to reassemble batches in the processor?

lib/racecar/multi_threaded_processor.rb
```diff
@@ -0,0 +1,215 @@
...
+          maybe_stop_or_exit(thread_key)
+          msgs = acquire_messages_from_queue(thread_key)
+          process_with_error_handling_and_retrying(thread_key, msgs) do
+            yield thread_key, msgs
+          end
+        end
+      end
+    end
+
+    def process_messages(thread_key, msgs)
```
dasch: I think having the message processing happen inline in the orchestrator makes the logic hard to follow; it also seems to me like Runner now _either_ manages message processing _or_ delegates to the orchestrator.

I think it would be better to extract a "processor" concept, perhaps as two classes: SingleThreadedProcessor and MultiThreadedProcessor. Instances of those would _always_ be tied to a single topic-partition, and would handle the relevant error handling etc. I don't like the Processing _module_, because mixing in stuff like that makes it much harder to follow.

That would also remove some complexity from Runner, which is a bit bloated.

Perhaps we could have the processors just have the `process` and `process_batch` methods, and the queueing would happen transparently in the multithreaded implementation?

```ruby
# MultiThreadedProcessor
def process_batch(message_batch)
  maybe_apply_backpressure
  @queue << message_batch
  # ...
end
```

Right now, "processor" refers to the client's `Racecar::Consumer` instance, but I think we could move things around a bit.

lib/racecar/cli.rb
```diff
@@ -65,8 +65,8 @@ def run
         config.install_liveness_probe
       end
 
-      processor = consumer_class.new
-      @runner = Racecar.runner(processor)
+      consumer_class_instance = consumer_class.new
```
dasch: I like `consumer_class_instance` :+1:

lib/racecar/multi_threaded_processor.rb
```diff
@@ -0,0 +1,217 @@
...
+    end
+
+    def initialize(config:, pauses:, consumer_class_instance:, instrumenter:, logger:)
+      @logger = logger
+      @threads = {}
+      @thread_queues = {}
+      @thread_mutexes = {}
+      @thread_metadata = {}
+      @finalize_mutex = Mutex.new
+      @config = config
```
dasch: I think we should have an instance of the processor _per topic-partition_, to avoid all this error-prone state management. So each instance would have a queue, mutex, metadata, etc.

ChrisGadek1: Right, I think the best idea here would be to delegate operations strictly related to threads into a dedicated class which I've named `ThreadManager`. Now this state looks cleaner is more more error-prone

lib/racecar/producer_methods.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Racecar
+  module ProducerMethods
+    def producer
+      @producer ||= Rdkafka::Config.new(producer_config).producer.tap do |producer|
+        producer.delivery_callback = Racecar::DeliveryCallback.new(instrumenter: @instrumenter)
+      end
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/consumer_set.rb
```diff
@@ -4,9 +4,10 @@ module Racecar
   class ConsumerSet
     MAX_POLL_TRIES = 10
 
-    def initialize(config, logger, instrumenter = NullInstrumenter)
+    def initialize(config, logger, instrumenter = NullInstrumenter, processor)
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/multi_threaded_processor.rb
```diff
@@ -0,0 +1,217 @@
+# frozen_string_literal: true
+
+require "racecar/processing"
+
+module Racecar
+  class MultiThreadedProcessor
+    include Processing
+    include ProducerMethods
+
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/multi_threaded_processor.rb
```diff
@@ -0,0 +1,217 @@
...
+          Thread.exit
+        else
+          Thread.stop
+        end
+      end
+    end
+
+    def maybe_resume_the_partition(thread_key, topic, partition)
+      if @thread_queues[thread_key].size < 0.5 * config.multithreaded_processing_max_queue_size
+        @finalize_mutex.synchronize do
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/pause.rb
```diff
@@ -2,7 +2,7 @@
 
 module Racecar
   class Pause
-    attr_reader :pauses_count
+    attr_reader :pauses_count, :backoff_interval
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,147 @@
+# frozen_string_literal: true
+
+require 'racecar/pause'
+
+module Racecar
+  class AsyncPartitionProcessor
+    attr_reader :thread, :queue, :config, :processor, :logger, :consumer, :consumer_class_instance, :instrumenter
+
+    THREAD_KEY = 'thread_key'.freeze
```
dasch: This seems like a poor choice of thread key

ChrisGadek1: Changed to
```
THREAD_KEY_IDENTIFIER = 'racecar_topic_partition_identifier'.freeze
```

lib/racecar/consumer_set.rb
```diff
@@ -49,13 +51,15 @@ def batch_poll(max_wait_time_ms = @config.max_wait_time_ms, max_messages = @conf
...
-        @logger.warn "Attempted to store_offset, but we're not subscribed to it: #{ErroneousStateError.new(e)}"
-        return
+      @runner_mutex.synchronize do
+        current.store_offset(message)
+      rescue Rdkafka::RdkafkaError => e
+        if e.code == :state # -172
+          @logger.warn "Attempted to store_offset, but we're not subscribed to it: #{ErroneousStateError.new(e)}"
+          return
+        end
+        raise e
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,147 @@
...
+          @logger.debug "Thread for #{thread_key} exiting"
+          Thread.exit
+        end
+      end
+    end
+
+    def maybe_apply_backpressure(messages)
+      if queue_size >= config.multithreaded_processing_max_queue_size
+        @runner_mutex.synchronize do
+          consumer.pause(@topic, @partition, Array(messages).last.offset + 1)
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+      end
+    end
+
+    def push(messages)
+      @mutex.synchronize do
+        @queue << Array(messages)
+        @cv.signal
+      end
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

spec/rebalance_listener_spec.rb
```diff
@@ -10,9 +10,17 @@
   let(:partitions_by_topic_hash) {
     { "topic_name" => [double(:partition, partition: 0, offset: 0, err: 0)] }
   }
+  let(:partition_processors) { { "topic_name/0": double(:processor) } }
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/consumer.rb
```diff
@@ -61,12 +71,16 @@ def teardown; end
...
     # caller must handle the latter cases or run into head of line blocking.
     def deliver!
-      @delivery_handles ||= []
-      if @delivery_handles.any?
-        instrumentation_payload = { delivered_message_count: @delivery_handles.size }
+      thread_key = Thread.current[AsyncPartitionProcessor::THREAD_KEY] || "main"
+      current_handles = maybe_synchronize do
+        @delivery_handles ||= {}
+        @delivery_handles[thread_key] ||= []
+      end
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+
+      spawn_thread do |msgs, use_process_batch|
+        if use_process_batch
+          processor.process_batch(msgs)
+        else
+          msgs.each do |msg|
+            processor.process(msg)
+          end
+        end
+      end
```
dasch: This conditional won't change dynamically; how about we instead decide up front which branch to use?


```suggestion
      if use_process_batch
        spawn_thread do |msgs|
          processor.process_batch(msgs)
        end
      else
        spawn_thread do |msgs|
          msgs.each do |msg|
            processor.process(msg)
          end
        end
      end
```

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+          loop do
+            wait_for_messages_or_exit
+            msgs = @queue.pop
+            @mutex.synchronize do
+              update_queue_size(msgs, operation: :subtract)
+            end
+            maybe_resume_the_partition
+            block.call(msgs, use_process_batch)
+          rescue => e
+            logger.error "Error in processing thread for #{thread_key}: #{e.class} - #{e.message}"
```
dasch: I think we should extract a `processing_loop` or similar private method to make it easier to understand the code; too much stuff is happening in this method.

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+        end
+      end
+    end
+
+    def push(messages)
+      @mutex.synchronize do
+        @queue << Array(messages)
+        update_queue_size(messages, operation: :add)
+        @cv.signal
+      end
```
dasch: I think we need a bit of analysis on this; I thought that `Queue` itself was thread safe, and so didn't need a mutex to protect writes to it?

Also, do we really need the separate queue size tracking? It has a `#length` method. What are the edge cases being solved here, if any?

dasch: I think having too many locks will lead to congestion and much lower performance; remember, this is called per message in some cases!

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+            Thread.exit
+          end
+          @cv.wait(@mutex)
+        end
+        if processor.rebalancing
+          @logger.debug "Thread for #{thread_key} exiting"
+          Thread.exit
+        end
+      end
+    end
```
dasch: Could we simplify or remove a lot of this be using Queue's non-blocking `#pop(true)` semantics? I would be okay with starting with this, but I think there needs to be some code comments explaining why this particular async model is used instead of a more simplistic Queue based one.

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    def maybe_apply_backpressure
+      if queue_size >= config.multithreaded_processing_max_queue_size
+        @runner_mutex.synchronize do
+          consumer.pause(@topic, @partition)
```
dasch: Is this the raw rdkafka consumer? Let's double check if that isn't thread safe already, it's pretty async in general.

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+        logger.debug "Paused partition #{@topic}/#{@partition}: queue reached capacity (#{queue_size}/#{config.multithreaded_processing_max_queue_size})"
+      end
+    end
+
+    def maybe_resume_the_partition
+      if queue_size < config.multithreaded_processing_resume_threshold * config.multithreaded_processing_max_queue_size
+        @runner_mutex.synchronize do
+          if consumer.respond_to?(:paused?)
+            return unless consumer.paused?(@topic, @partition)
+          end
```
dasch: What's going on here? Why would the consumer not have that method?

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,168 @@
...
+          @queue_size += 1
+        end
+      elsif operation == :subtract
+        if config.multithreaded_processing_queue_size_by_message_count
+          @queue_size -= Array(messages).size
+        else
+          @queue_size -= 1
+        end
+      end
+    end
```
dasch: This seems like a much more complicated way to do `queue.length`

lib/racecar/consumer.rb
```diff
@@ -61,12 +71,16 @@ def teardown; end
...
-      @delivery_handles ||= []
-      if @delivery_handles.any?
-        instrumentation_payload = { delivered_message_count: @delivery_handles.size }
+      thread_key = Thread.current[AsyncPartitionProcessor::THREAD_KEY_IDENTIFIER] || Thread.current.object_id
+      current_handles = maybe_synchronize do
+        @delivery_handles ||= {}
+        @delivery_handles[thread_key] ||= []
+      end
+      if current_handles.any?
+        instrumentation_payload = { delivered_message_count: current_handles.size }
```
dasch: This whole thing feels wrong; when in async mode, the consumer class instances live inside a single thread within the async processor instances. Why do we need thread safe handling here? I think this is a leftover from the previous iteration. It also shows how much complexity was introduced by that. We should have the instances keep track of the handles of the messages they _themselves_ produce, and `#deliver!` should be scoped to those handles – so I think the code in master should work out of the box probably?

lib/racecar/consumer_set.rb
```diff
@@ -49,13 +51,22 @@ def batch_poll(max_wait_time_ms = @config.max_wait_time_ms, max_messages = @conf
...
+        end
+      rescue Rdkafka::RdkafkaError => e
+        if e.code == :state # -172
+          @logger.warn "Attempted to store_offset, but we're not subscribed to it: #{ErroneousStateError.new(e)}"
+          return
+        end
+        raise e
       end
-      raise e
     end
```
dasch: When is this method called? It seems like this should maybe be moved into the processors, which already have the relevant `consumer`.

dasch (changes requested):
I think in general we need to have a more clear boundary between single-threaded and multi-threaded code; I think that's the async processor. Anything "inside" such a processor should assume to be running in a single thread. Maybe there's a couple exceptions, but in general that should be the design principle. I don't think we should have mutexes smattered all over the code.

lib/racecar/producer_methods.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "rdkafka"
+require "racecar/delivery_callback"
+
+module Racecar
+  module ProducerMethods
```
dasch: Why do we need this module?

lib/racecar/async_partition_processor.rb
ChrisGadek1: Copilot stated that previous `queue.length` was returning the number of messaging batches (and not the actual number of messages in the queue). And yes, `Queue` is thread safe and can be moved outside the synchronisation

lib/racecar/processing.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module Racecar
+  module Processing
```
dasch: Why do we need this module?

lib/racecar/async_partition_processor.rb
ChrisGadek1: It's `ConsumerSet` instance

lib/racecar/consumer.rb
ChrisGadek1: Right now we're sharing the same user-defined consumer instance between all processor, but that seems wrong. I'll try to assign separate instances into processors

lib/racecar/pause.rb
```diff
@@ -4,6 +4,28 @@ module Racecar
...
+      Hash.new {|h, k|
+        h[k] = Hash.new {|h2, k2|
+          h2[k2] = ::Racecar::Pause.new(
+            timeout:             timeout,
+            max_timeout:         config.max_pause_timeout,
+            exponential_backoff: config.pause_with_exponential_backoff
+          )
+        }
+      }
+    end
```
dasch: This seems to generate a two-level hash for keeping track of pauses across _all_ the topic-partitions; yet when used from AsyncPartitionProcessor we should only ever care about a single topic-partition at a time. I think we should just use `Pause.new` there? I'm not sure I see the need for this method. If we need to, perhaps move the normalization and validation of `pause_timeout` into the initializer.

lib/racecar/consumer_set.rb
```diff
@@ -92,38 +96,47 @@ def each_subscribed
       end
     end
 
-    def pause(topic, partition, offset)
+    def pause(topic, partition, offset = nil)
```
dasch, line 115: I'm trying to understand why this method needs to become more complicated; was it because we're still sharing an instance of the class between all processors?

I think the whole point of having the processors is to have thread-local homes for topic-partition specific state.

Try to look into the state stored in `@paused_tpls`; could we instead store it in the processors? Then it would just be a flat ivar, because each processor only deals with one topic-partition. And this method (and `#resume`) would just provide an API to call the underlying rdkafka consumer.

We might even want to go further; if Processor handles topic-partition specific state, and we currently have rdkafka consumer instances per topic-partition, then maybe instead of having ConsumerSet we should store those rdkafka consumers in the processors? (Let's just call them `rdkafka_consumer` internally to avoid confusion.)

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,221 @@
...
+require "rdkafka"
+require "racecar/pause"
+require "racecar/delivery_callback"
+
+module Racecar
+  class PartitionProcessor
+    attr_reader :consumer_class_instance, :config, :logger, :instrumenter, :consumer, :pauses
+    attr_accessor :rebalancing, :shutting_down
+
+    def initialize(config:, logger:, instrumenter:, consumer_class:, consumer:, pauses:)
```
dasch: Should this class not know the topic and partition it's processing?

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,221 @@
+# frozen_string_literal: true
+
+require "rdkafka"
+require "racecar/pause"
+require "racecar/delivery_callback"
+
+module Racecar
+  class PartitionProcessor
```
dasch, line 8: Let's document the purpose of the classes, it also helps the AI be more effective.

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,221 @@
...
+      producer.close
+    end
+
+    private
+
+    def resume_all_paused_partitions
+      return if config.pause_timeout == 0
+
+      pauses.each do |topic, partitions|
+        partitions.each do |partition, pause|
```
dasch: We still seem to be working _across_ topic-partitions here, but I thought the whole point of processors was to keep all topic-partition state localized in a single instance!

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,124 @@
...
+        @backpressure_paused.make_true
+        consumer.pause(@topic, @partition)
+        logger.debug "Paused partition #{@topic}/#{@partition}: queue reached capacity (#{@queue.size}/#{config.multithreaded_processing_max_queue_size})"
+      end
+    end
+
+    def maybe_resume_the_partition
+      if @backpressure_paused.true? && @queue.size < config.multithreaded_processing_resume_threshold * config.multithreaded_processing_max_queue_size
+        @backpressure_paused.make_false
+        consumer.resume(@topic, @partition)
```
dasch: Can this fail, i.e. could we have a discrepancy between `@backpressure_paused` and the consumer? Same goes for the backpressure method.

Is there a way to query `consumer` to see if the partition is paused, and would that perhaps be just as fast as the atomic boolean? Then we'd delegate the state to `consumer` and don't need to worry about integrity.

lib/racecar/consumer.rb
```diff
@@ -91,7 +89,8 @@ def deliver!
           end
         end
       end
-      @delivery_handles.clear
+    ensure
+      current_handles&.clear
```
dasch: I'm not sure what this change does; we'd still be destructively cleaning `@delivery_handles` even if we assign the array to `current_handles`.

lib/racecar.rb
```diff
@@ -65,12 +65,12 @@ def self.instrumenter
...
+  def self.run(consumer_class)
+    runner(consumer_class).run
   end
 
-  def self.runner(processor)
-    runner = Runner.new(processor, config: config, logger: logger, instrumenter: config.instrumenter)
+  def self.runner(consumer_class)
+    runner = Runner.new(consumer_class, config: config, logger: logger, instrumenter: config.instrumenter)
 
     if config.parallel_workers && config.parallel_workers > 1
```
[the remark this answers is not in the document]
dasch: That should not be a problem

lib/racecar/async_partition_processor.rb
ChrisGadek1: Unfortunately, Rdkafka has no exposed API to check if a consumer is paused. But moving it into `ConsumerSet` would need to introduce a set of pauses there and we just removed it for simplicity.

lib/racecar/consumer.rb
ChrisGadek1: This should removed. Already done

lib/racecar/consumer_set.rb
```diff
@@ -49,7 +52,8 @@ def batch_poll(max_wait_time_ms = @config.max_wait_time_ms, max_messages = @conf
     end
 
     def store_offset(message)
-      current.store_offset(message)
+      found_consumer, _ = find_consumer_by(message.topic, message.partition)
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/consumer_set.rb
```diff
@@ -66,13 +70,28 @@ def commit
 
     def close
       each_subscribed(&:close)
-      @paused_tpls.clear
+      @producer&.close
+      @producer = nil
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/runner.rb
```diff
@@ -166,128 +124,66 @@ def install_signal_handlers
...
-        desc = "#{topic}/#{partition}"
-        logger.error "Failed to process #{desc} at #{offsets}: #{e}"
-
-        logger.warn "Pausing partition #{desc} for #{pause.backoff_interval} seconds"
-        consumer.pause(topic, partition, offsets.first)
-        pause.pause!
+        @consumer_class_instance.deliver!
+      ensure
+        @consumer_class_instance.teardown
       end
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/rebalance_listener.rb
```diff
@@ -24,6 +27,18 @@ def on_partitions_revoked(rdkafka_topic_partition_list)
...
+        rdkafka_topic_partition_list.to_h.each do |topic, partitions_metadata|
+          partitions_metadata.flatten.map(&:partition).each do |partition|
+            key = Runner.topic_partition_key(topic, partition)
+            @runner_mutex.synchronize do
+              processor = @partition_processors[key]
+              processor&.rebalancing = true
+              processor&.resume_paused_partition if processor.respond_to?(:resume_paused_partition)
+              @partition_processors.delete(key)
+
+            end
```
dasch: This code seems a bit off to me:
* We should not have a different API for resuming paused partitions between the sync and async processor class, I think?
* Setting the attribute directly on the instance seems wrong; I think we should instead do something like `processor.revoke_partition!` or something, and have the processor (sync or async) handle the state and cleanup internally.
* We're deleting the processor from the `@partition_processors`, but are we also shutting down the internal thread and all that? Probably the method above should take care to clean up all the resources managed by the processor.

lib/racecar/runner.rb
```diff
@@ -166,128 +124,67 @@ def install_signal_handlers
...
-          instrumentation_payload[:unrecoverable_delivery_error] = reset_producer_on_unrecoverable_delivery_errors(e)
-          instrumentation_payload[:retries_count] = pause.pauses_count
-          config.error_handler.call(e, instrumentation_payload)
-          raise e
-        end
+    def assign_and_get_processor(messages)
+      topic     = messages.is_a?(Array) ? messages.first.topic     : messages.topic
+      partition = messages.is_a?(Array) ? messages.first.partition : messages.partition
+      key = Runner.topic_partition_key(topic, partition)
+      @mutex.synchronize do
```
dasch: This mutex is problematic; it is locked on _every batch_ or _every message_, meaning tens to thousands of times for each Kafka fetch response. I think we should be able to use a lock-free hash to avoid congestion on reads, and only use a mutex when _writing_ to it, i.e. when we are lazily constructing a new processor.

lib/racecar/runner.rb
```diff
@@ -6,156 +6,114 @@
...
+
+          @instrumenter.instrument("start_main_loop", loop_payload)
+          @instrumenter.instrument("main_loop", loop_payload) do
+            resume_all_paused_partitions unless config.multithreaded_processing_enabled
+
+            case process_method
+            when :batch then
+              msg_per_part = consumer.batch_poll(config.max_wait_time_ms).group_by(&:partition)
+              msg_per_part.each_value do |messages_per_partition|
+                messages_per_partition.group_by(&:topic).each_value do |messages_per_topic_and_partition|
```
dasch: Do we need to group by topic here? My understanding is that `partition` is the topic-partition, i.e. encodes both the topic and partition number, and we should just be able to to `topic = messages_per_partition.first.topic` or something. This is a hot loop, so we want to avoid unnecessary allocations.

lib/racecar/runner.rb
```diff
@@ -166,128 +124,67 @@ def install_signal_handlers
...
-      @instrumenter.instrument("start_process_batch", instrumentation_payload)
-      with_pause(first.topic, first.partition, first.offset..last.offset) do |pause|
-        begin
-          @instrumenter.instrument("process_batch", instrumentation_payload) do
-            racecar_messages = messages.map do |message|
-              Racecar::Message.new(message, retries_count: pause.pauses_count)
+    def shutdown_processors_and_wait
+      if config.multithreaded_processing_enabled
+        processors_snapshot = @mutex.synchronize { partition_processors.values }
+        processors_snapshot.each { |processor| processor.shutting_down = true if processor }
```
dasch: I don't like using attribute writers for this; let's instead use `shut_down!` or something.

dasch (changes requested):

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+require 'racecar/pause'
+require 'concurrent-ruby'
+
+module Racecar
+  class AsyncPartitionProcessor
```
dasch, line 7: I think it would be nice to document the new classes to explain their purpose.

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,131 @@
...
+    attr_reader :thread, :queue, :config, :processor, :logger, :consumer, :consumer_class, :instrumenter, :backpressure_paused
```
dasch: Do these need to be read outside of the class? Could we make them private perhaps?

lib/racecar/runner.rb
```diff
@@ -166,128 +131,68 @@ def install_signal_handlers
...
+
+      if config.multithreaded_processing_enabled
+        @mutex.synchronize do
+          processor = AsyncPartitionProcessor.new(
+            **common_processor_params,
+            consumer_class: consumer_class,
+            topic: topic,
+            partition: partition,
+          )
+          partition_processors[key] = processor
```
dasch: Do we need a mutex now that we're using a concurrent hash to those the processors? I would think not.

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,210 @@
...
+        logger.info "Automatically resuming partition #{topic}/#{partition}, pause timeout expired"
+        consumer.resume(topic, partition)
+        pause.resume!
+      end
+    end
+
+    def rebalance!
+      @rebalancing = true
+      @sleep_mutex.synchronize { @sleep_cv.signal }
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Very unlikely to happen and trying to resume there may lead to error, because the partition is already revoked, so it cannot be resumed from here

lib/racecar/runner.rb
ChrisGadek1: Removed 👍

lib/racecar/async_partition_processor.rb
ChrisGadek1: Most of them moved to `private` section

lib/racecar/async_partition_processor.rb
ChrisGadek1: I have a plan to add documentation after finishing testing, before final polishing

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,210 @@
...
+          if rebalancing
+            Thread.exit
+          elsif !shutting_down
+            handle_processing_error(e, payload, pause: pause)
+            pause.pause!
+            unless config.pause_timeout <= 0
+              @sleep_mutex.synchronize do
+                @sleep_cv.wait(@sleep_mutex, pause.backoff_interval)
+              end
+            end
```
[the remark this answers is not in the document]
ChrisGadek1: If I understand it correctly, this is the same code (in the suggestion)

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    def rebalance!
+      processor.rebalance!
+      @queue << nil
+    end
+
+    def shut_down!
+      processor.shut_down!
+      @queue << nil
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: After rebalance or shutdown the worker thread is exited and async processor is removed, so this scenario will not happen

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,210 @@
...
+          consumer.store_offset(messages.last) unless rebalancing
+        end
+      end
+    end
+
+    def teardown
+      consumer_class_instance.deliver! unless rebalancing
+    ensure
+      consumer_class_instance.teardown
+    end
```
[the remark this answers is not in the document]
ChrisGadek1: Updated description

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,131 @@
...
+      @queue << Array(messages)
+      maybe_apply_backpressure
+    end
+
+    def maybe_apply_backpressure
+      if @queue.size >= config.multithreaded_processing_max_queue_size
+        @backpressure_paused.make_true
+        consumer.pause(@topic, @partition)
+        logger.debug "Paused partition #{@topic}/#{@partition}: queue reached capacity (#{@queue.size}/#{config.multithreaded_processing_max_queue_size})"
+      end
```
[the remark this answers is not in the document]
ChrisGadek1: added suggested guard

lib/racecar/consumer_set.rb
```diff
@@ -66,13 +75,29 @@ def commit
 
     def close
       each_subscribed(&:close)
-      @paused_tpls.clear
+      @producer&.close
+      @producer = nil
```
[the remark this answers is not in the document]
ChrisGadek1: Fixed

lib/racecar/rebalance_listener.rb
```diff
@@ -24,6 +27,15 @@ def on_partitions_revoked(rdkafka_topic_partition_list)
 
       instrument("partitions_revoked", partitions: event.partition_numbers) do
         consumer_class.on_partitions_revoked(event)
+        rdkafka_topic_partition_list.to_h.each do |topic, partitions_metadata|
+          partitions_metadata.flatten.map(&:partition).each do |partition|
+            key = Runner.topic_partition_key(topic, partition)
+            @runner_mutex.synchronize do
+              processor = @partition_processors[key]
+              processor&.rebalance!
+            end
```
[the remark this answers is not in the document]
ChrisGadek1: At this point the partition could be already revoked, so trying to unpause it may raise error

dasch (dismissed):

dasch (dismissed):

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+require 'racecar/pause'
+require 'concurrent-ruby'
```
batizhevsky, line 4: I think you need to add it to `add_runtime_dependency` in `gemspec` file

ChrisGadek1: Sure, added

batizhevsky (dismissed):

lib/racecar/partition_processor.rb
```diff
@@ -0,0 +1,220 @@
...
+            handle_processing_error(e, payload, pause: pause)
+            pause.pause!
+            unless config.pause_timeout <= 0
+              @sleep_mutex.synchronize do
+                next if rebalancing || shutting_down
+                @sleep_cv.wait(@sleep_mutex, pause.backoff_interval)
+              end
+            end
+            Thread.exit if rebalancing
+            break if shutting_down || config.pause_timeout <= 0
```
Thomascountz: When `pause_timeout = -1` this skips the failed message and commits past it, causing silent message loss.

`-1` means "pause indefinitely", i.e., at-least-once and strict ordering, according to the single-threaded implementation. In this implementation, the guard on `:145` will skip the wait and the guard on `:152` will `break` out of the loop and the `process` method returns to `main_processing_loop` without retrying, rewinding, or pausing the partition in rdkakfa. The `main_processing_loop`'s `loop` continues to run, it picks up on the next message in the queue, and if that success, we end up calling `consumer.store_offset`. The failed message ends up being silently dropped since offset will now skip ahead.

lib/racecar/async_partition_processor.rb
```diff
@@ -0,0 +1,138 @@
...
+      push(message)
+    end
+
+    def process_batch(messages)
+      push(messages)
+    end
+
+    def rebalance!
+      processor.rebalance!
+      @queue << nil
```
Thomascountz, line 42: From what I understand, if an async worker is "rebalanced," the queue will keep being drained until it hits this `nil` sentinel, and during that time, side effects are still run, just without advancing the offset counter. 

Racecar will process messages it knows cannot commit and the old member can still run application code for already-fetched messages while the new owner may start processing from the last committed offset.

I'm wondering why we don't let the one in-flight message finish, then stop without picking up new work?

lib/racecar/runner.rb
```diff
@@ -166,128 +135,65 @@ def install_signal_handlers
...
+      if config.multithreaded_processing_enabled
+        processors_snapshot = partition_processors.values
+        processors_snapshot.each { |processor| processor.shut_down! if processor }
+        processors_snapshot.each do |processor|
+          if processor.respond_to?(:thread)
+            begin
+              processor.thread.join(config.multithreaded_processing_shutdown_timeout)
+            rescue => e
+              logger.error "Error while waiting for processor thread to finish: #{e}"
             end
```
Thomascountz, line 177: `Thread#join(timeout)` returns `nil` on timeout, rather than raising. These threads are abandoned and may continue making noise. What do you think about checking for `nil` and adding a log for visibility into threads that timeout during shutdown. Because we still have `process.thread`, we could log its metadata as well as the error.

lib/racecar/partition_processor.rb
ChrisGadek1: I've fixed that wrong behaviour. Right now the pause is indefinite if the timeout is set to `-1` and no message is dropped silently

lib/racecar/async_partition_processor.rb
ChrisGadek1: That's a good catch, I was trying to mitigate it by trying to add the sentinel at the beginning of the queue, but it's not possible for `Queue.new`. So, instead I've added the explicit method `exit_if_rebalancing!` addressing the issue you mentioned.

lib/racecar/runner.rb
ChrisGadek1: That's also a good catch, I've fixed it by raising (and catching) an error when the timeout occurs. I think the abandoned threads will not live long after the timeout, because this situation could only occur during graceful finish of the Racecar consumer, but it's worth to log it.

PR #415, opened 2026-06-30
[Roll out of 3.0.0 Part 1] Add Partition Processor and fix flakky tests

ChrisGadek1:
## What Changes

Extract message processing and instrumentation into PartitionProcessor; fix flakky test
Move the per-message/per-batch processing and its instrumentation out of
Runner and into a dedicated PartitionProcessor:

- PartitionProcessor builds the instrumentation payloads (message_payload/
  batch_payload), wraps processing in the process_message/process_batch
  instrumentation, runs the consumer's process/process_batch, flushes the
  producer and stores the offset.
- Runner keeps a map of PartitionProcessors keyed by topic/partition
  (assign_and_get_processor) and delegates the actual processing to them,
  while retaining ownership of pausing and the producer-reset error
  handling.

Also, tests in ./spec/integration/cooperative_sticky_assignment_spec.rb are flakky,
so this commit contains also a fix for them

Pausing, error handling and producer lifecycle stay in Runner for now.
Behaviour-preserving for the single-threaded path. All unit specs pass.

## Breaking Changes

* No

### DIFF
--- a/lib/racecar/runner.rb
+++ b/lib/racecar/runner.rb
@@ -6,6 +6,7 @@ require "racecar/message"
 require "racecar/message_delivery_error"
 require "racecar/erroneous_state_error"
 require "racecar/delivery_callback"
+require "racecar/partition_processor"
 
 module Racecar
   class Runner
@@ -15,6 +16,7 @@ module Racecar
       @processor, @config, @logger = processor, config, logger
       @instrumenter = instrumenter
       @stop_requested = false
+      @partition_processors = Hash.new
       Rdkafka::Config.logger = logger
 
       if processor.respond_to?(:statistics_callback)
@@ -22,6 +24,10 @@ module Racecar
       end
 
       setup_pauses
+    end
+
+    def self.topic_partition_key(topic, partition)
+      "#{topic}/#{partition}"
     end
 
     def setup_pauses
@@ -135,6 +141,15 @@ module Racecar
       end
     end
 
+    def assign_and_get_processor(messages)
+      topic     = messages.is_a?(Array) ? messages.first.topic     : messages.topic
+      partition = messages.is_a?(Array) ? messages.first.partition : messages.partition
+      key = Runner.topic_partition_key(topic, partition)
+      return @partition_processors[key] if @partition_processors[key]
+
+      @partition_processors[key] = PartitionProcessor.new(processor: processor, consumer: consumer, instrumenter: @instrumenter)
+    end
+
     def producer
       @producer ||= Rdkafka::Config.new(producer_config).producer.tap do |producer|
         producer.delivery_callback = Racecar::DeliveryCallback.new(instrumenter: @instrumenter)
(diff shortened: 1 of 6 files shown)


lib/racecar/runner.rb
```diff
@@ -167,25 +182,13 @@ def install_signal_handlers
...
-        consumer_class: processor.class.to_s,
-        topic:          message.topic,
-        partition:      message.partition,
-        offset:         message.offset,
-        create_time:    message.timestamp,
-        key:            message.key,
-        value:          message.payload,
-        headers:        message.headers
-      }
+      processor = assign_and_get_processor(message)
```
HeyNonster, line 185: Should we just be passing in the topic and partition instead of a message here? It looks like we're instantiating a processor with a message but in actuality we're just using the topic/partition of the message to return or instantiate a processor.

lib/racecar/runner.rb
```diff
@@ -167,25 +182,13 @@ def install_signal_handlers
...
 
       @instrumenter.instrument("start_process_message", instrumentation_payload)
       with_pause(message.topic, message.partition, message.offset..message.offset) do |pause|
         begin
-          @instrumenter.instrument("process_message", instrumentation_payload) do
-            processor.process(Racecar::Message.new(message, retries_count: pause.pauses_count))
-            processor.deliver!
-            consumer.store_offset(message)
-          end
+          processor.process(message, pause.pauses_count, instrumentation_payload)
```
HeyNonster, line 191: It feels like some of this was half-extracted. i.e. We ask the processor for an `instrumentation_payload` here then pass that same payload provided by the processor back to itself. It appears that was done because we need the payload for error handling below but perhaps the error handling should live in the partition processor if that's the case?

lib/racecar/runner.rb
```diff
@@ -196,28 +199,14 @@ def process(message)
...
-      instrumentation_payload = {
-        consumer_class: processor.class.to_s,
-        topic:          first.topic,
-        partition:      first.partition,
-        first_offset:   first.offset,
-        last_offset:    last.offset,
-        last_create_time: last.timestamp,
-        message_count:  messages.size
-      }
+      processor = assign_and_get_processor(messages)
```
HeyNonster, line 202: If we're passing the topic/partition to `assign_and_get_processor` instead of a message or (possibly) messages then it would likely simplify the `assign_and_get_processor` method. Instead of checking the argument it takes is an array or a single object, maybe `process` and `process_batch` handle figuring out what topic/partition to provide it.

lib/racecar/runner.rb
```diff
@@ -135,6 +141,15 @@ def consumer
       end
     end
 
+    def assign_and_get_processor(messages)
+      topic     = messages.is_a?(Array) ? messages.first.topic     : messages.topic
+      partition = messages.is_a?(Array) ? messages.first.partition : messages.partition
+      key = Runner.topic_partition_key(topic, partition)
+      return @partition_processors[key] if @partition_processors[key]
+
+      @partition_processors[key] = PartitionProcessor.new(processor: processor, consumer: consumer, instrumenter: @instrumenter)
```
HeyNonster, line 150:
```suggestion
@partition_processors[key] ||= PartitionProcessor.new(processor: processor, consumer: consumer, instrumenter: @instrumenter)
```

spec/integration/cooperative_sticky_assignment_spec.rb
HeyNonster, line 1: Are these flaky on main? If so, let's fix these separately.

HeyNonster:
Also, conceptually, it would be useful for us to have more explanation of the why in the PR description. We have some of the context from the larger PR but it would be helpful to have an outline of the steps you intend to take to get from where we are today to the final goal of an async processor.

As step one, what is a Partition Processor and why is it necessary? What does this unlock for step 2? What does step 3, 4, 5, etc look like after this? :)

HeyNonster:
Also, we (@zendesk/ruby-core) requested these smaller, incremental steps and we appreciate the work that you're doing here in regards to that endeavor!

